PR #9, opened 2017-03-23
Allow enums to be strings, keywords, or symbols

### LINKED ISSUE #1: Enums can be defined as symbols (or keywords) causing inscrutable failures
It's is possible to define the :values of an enum as symbols; this doesn't give a compile time or spec error BUT at runtime, request enum values (as strings) will not compare against schema enum values (as symbols). This is further obscured when the resulting error is reported as JSON instead of EDN, as the symbols in the :enum-values key are converted to strings.

hlship:
Fixes #1 
 
@annapawlicka @bcarrell

### DIFF
--- a/docs/enums.rst
+++ b/docs/enums.rst
@@ -11,6 +11,13 @@ values are restricted to a predefined set.
 .. literalinclude:: _examples/enum-definition.edn
    :language: clojure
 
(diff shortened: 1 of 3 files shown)


annapawlicka:
LGTM 👍

PR #10, opened 2017-03-24
A few spot optimizations

mpenet:
It seems the :else clause will just return nil if (nil? x) is true anyway. 

The 2 other commits are just performance enhancements (without making the code uglier (imho)).
Thanks for the lib, it's great stuff.

### DIFF
--- a/src/com/walmartlabs/lacinia/internal_utils.clj
+++ b/src/com/walmartlabs/lacinia/internal_utils.clj
@@ -37,7 +37,7 @@
 
 (defn keepv
   [f coll]
-  (into [] (keep f coll)))
+  (into [] (keep f) coll))
 
 (defn hash-map-by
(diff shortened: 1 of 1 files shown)


hlship:
LGTM

In general, we'd prefer that these kind of optimizations be accompanied by benchmarks to show they are worthwhile; these are tiny and isolated enough that it doesn't matter, but if the changes affected multiple namespaces, we'd want to offset the risk by showing a clear benefit.  See the `perf` namespace and the `lein benchmarks` alias.

bcarrell:
Great, thank you!  I'll echo @hlship and suggest that including a few benchmarks (eg. `criterium.core/bench`) of these fns on an individual scale would be helpful.  In any event, we verified the performance improvements and readability is equivalent so it looks like a win to me!

Thanks a lot!  🌟

PR #12, opened 2017-03-24
Add new benchmarks and generate performance graphs

hlship:
I'm starting to accumulate new data here, which is nice; it already includes mpenet's optimizations (which didn't affect actual parse or execution time).

New benchmarks:
- A more modest simple query
- The simple query, but using variables

exec-time
parse-time

@annapawlicka @bcarrell


dev-resources/perf.clj
```diff
@@ -85,47 +94,110 @@
...
+   :basic
+   {:query "{
+   default: human { name appears_in friends { name } home_planet }
+   hope_hero: hero(episode: NEWHOPE) { id name friends { name }}
+   }"}
+   :basic-vars
+   {:query "query ($ep : episode!) {
+   default: human { name appears_in friends { name } home_planet }
+   hope_hero: hero(episode: $ep) { id name friends { name }}
+   }"
```
annapawlicka: Could we re-format the queries for better readability?

dev-resources/perf.clj
```diff
@@ -85,47 +94,110 @@
...
+      str/trim
+      (subs 0 8)))
+
+(def ^:private dataset-file "perf/benchmarks.csv")
+
+(defn run-benchmarks [options]
+  (let [prefix [(format "%tY%<tm%<td" (Date.))
+                (-> options
+                    :commit
+                    (or (git-commit)))]
```
annapawlicka: This would read better as `(or (:commit options) (git-commit))`

dev-resources/perf.clj
```diff
@@ -85,47 +94,110 @@
...
+        new-benchmarks (->> (map run-benchmark (keys benchmark-queries))
+                            (map #(into prefix %)))
+        dataset (-> (read-dataset dataset-file  :header true)
```
annapawlicka: Extra space

Issue #2, opened 2017-03-18
Error Messages -- NPE on fragment w/ undefined type


eyston:
Howdy!

This isn't a bug, just a quality of life improvement on error messaging :)

If you query a fragment w/ an undefined type you get an NPE on this line:

https://github.com/walmartlabs/lacinia/blob/30da50a32eb5772bea81914d0988f4370158a70a/src/com/walmartlabs/lacinia/parser.clj#L753

The `condition-type` is `nil`:

https://github.com/walmartlabs/lacinia/blob/30da50a32eb5772bea81914d0988f4370158a70a/src/com/walmartlabs/lacinia/parser.clj#L800

Can reproduce using the sample schema:

```clj
(parser/parse-query schema "query { human(id: \"1001\"){ ... on foo { name }}}")
;; CompilerException java.lang.NullPointerException
```

I would expect something similar as you've done for undefined fields:

```clj
(parser/parse-query schema "query { human(id: \"1001\"){ foo }}")
;; CompilerException clojure.lang.ExceptionInfo: Cannot query field `foo' on type `human'. {:query-path [:human], :locations [{:line 1, :column 25}], :field :foo, :type :human}
```

Thanks for releasing this -- I look forward to when you can accept contributions :)

bcarrell:
Thanks for the report!  We'll take a look at this soon.

PR #13, opened 2017-03-24
Throw exception for inline fragments with invalid type conditions.

annapawlicka:
If inline fragment contains type condition, we should check if this type is defined in schema. At the moment we're throwing NPE. This PR fixes https://github.com/walmartlabs/lacinia/issues/2.

Ping @hlship @bcarrell


hlship:
LGTM

PR #14, opened 2017-03-24
Refactoring of execution

hlship:
This is a number of refactorings of the execution namespace, to get ready for async query processing.

I decided to stop before the PR got too big (for once).

Part of this simplifies and improves error reporting, at the cost of being less functional (an Atom is passed around).

This builds on my other PR, which should be committed first.

@annapawlicka @sashton


perf/benchmarks.csv
```diff
@@ -18,3 +18,7 @@ date,commit,kind,parse,exec
 20170324,1b7f456c,introspection,2.0246591950333332,3.4602085544289043
 20170324,1b7f456c,basic,0.5015100832206664,0.24973527992753625
 20170324,1b7f456c,basic-vars,0.6428995111702128,0.25122130674037674
+20170324,cae8f794,introspection,2.0250440356182797,3.178506137552083
+20170324,cae8f794,basic,0.4980364009556764,0.24132741184534814
+20170324,cae8f794,basic-vars,0.6358850933709745,0.24591709582839535
+20170324,cae8f794,errors,0.23813962460006155,6.3390120183333325
```
hlship: Don't have that, though with some effort I could do a little time travel to gather that.
The reality of the change is to simplify the code paths before introducing the async stuff, which will make it all more complicated.
Basically, we can now handle errors quickly at the junction between application resolver functions and the rest of Lacinia.

hlship: Some minor perf. improvement, too:  With the old, non-atom approach, I'm seeing 0.24 parse and 6.9 exec.  Pretty marginal.

What's the right term for "don't know if this is accurate, or just random noise"?

bcarrell:
I'll take a look at this tomorrow morning but if @sashton gets to it before me, feel free. :)

bcarrell:
I think I'm OK with this, even though adding an atom here feels not great.  I can see how it would be useful for making future adjustments.  If you remember, we used an atom to collect errors in the earlier versions.  If you're comfortable with it, I'm comfortable with it.

hlship:
Please merge. I think this is the right approach, as it simplifies much that I would prefer straightened out prior to taking a crack at proper types.

PR #22, opened 2017-03-29
readme improvements

am0ry:
A few README improvements and an attempt to make it read more like a tutorial.


am0ry:
Is there a tutorial available in the manual? I couldn't find one. I think it's nice to have a working example that's easily discoverable, either in the README or linked from it to help out anyone evaluating this library.

bcarrell:
Merging this; we do need to get a fuller example/tutorial going, though.

Issue #21, opened 2017-03-28
Accept clojure sets as valid graphql list type when returned from resolver


mpenet:
Right now it returns sets as strings. 
This is something quite common in clojure apis and would be more friendly than having to walk our return data to convert sets to list/vec or define a custom scalar that reads the string (lots of wasted cycles for something like this, set->string->set). Worth noting that's something all (?) json marshaling libraries for clojure do transparently too.

it seems mostly due to` sequential?` checks instead of `(or (sequential? x) (set? x))`, ex here https://github.com/walmartlabs/lacinia/blob/61bcaa20cacc66ebbb20b8adf07949fd45d124ef/src/com/walmartlabs/lacinia/schema.clj#L296-L308

bcarrell:
Seems fine to me, feel free to send a PR or we'll work on fixing this soon!

mpenet:
No longer relevant since #23 was merged

PR #23, opened 2017-03-29
fix condition checking only sequential for collection like things 

mpenet:
This allow Sets to be used as field values in resolver returned data. Before this patch it errors on a sequential? check and thus disallows the use of Set. This is similar to how json encoders operate (encoding Sets as Arrays).
related to #21

### DIFF
--- a/src/com/walmartlabs/lacinia/internal_utils.clj
+++ b/src/com/walmartlabs/lacinia/internal_utils.clj
@@ -135,6 +135,12 @@
   [v]
   (str \` (name v) \'))
 
+(defn sequential-or-set?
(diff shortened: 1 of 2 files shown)


bcarrell:
Thanks!  Merged this one manually (after removing a stray newline).

mpenet:
Thanks, any chance of a quick release including this fix? Otherwise we'd just use a private version in the meantime.

bcarrell:
I don't see why not.  @hlship ^^

hlship:
I can do a release at any time.  I've been wanting to do a release, and base the lacinia-pedestal library on that dot release (due to the changes in execute).

I can do a release after the morning meeting.

PR #24, opened 2017-03-29
Further spot performance improvements

mpenet:
Small gains probably, I didn't measure yet. 
Even if some of the modifications do not show a lot of gains in pure speed it should be more "garbage friendly". 
This is just "surface" work, I am sure there are more gains to be had by changing how things are done in some places, but I am not so familiar with the codebase yet nor do I have a lot of free time at the moment :baby_bottle: :baby:

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -96,7 +96,7 @@
 (defn ^:private coerce-to-int
   [v]
   (cond
-    (number? v) (.intValue v)
+    (number? v) (.intValue ^Number v)
     (string? v) (Integer/parseInt v)
     :else (throw (ex-info (str "Invalid Int value: " v) {:value v}))))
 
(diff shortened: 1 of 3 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -445,12 +445,11 @@ z
           resolver-errors (when is-tuple?
                             (resolve-errors raw-value))
           enforced-tuple (enforcer [enforce-value])
-          all-errors (->> (concat
-                            (ensure-seq resolver-errors)
-                            (ensure-seq (second enforced-tuple)))
-                          (filter some?)
-                          (mapcat assert-and-wrap-error)
-                          seq)]
+          all-errors (->> (sequence (comp cat
```
mpenet: might be faster with `into`, depending if we care about laziness

bcarrell:
Appreciate you taking a look at possibilities for improving performance -- I'll try and add some measurements.

hlship:
I'm running a check on performance right now.  Getting rid of some reflection during parsing could show up.

hlship:
Big drop in parse time.
parse-time
Small up-tick in execution time:

exec-time

bcarrell:
Looks good to me then.  I'm fine with merging this.

hlship:
I've merged this, along with updated benchmarks.

PR #28, opened 2017-04-03
Load clojure.set dependency

wilkerlucio:
Without this (and unless you have loaded `clojure.set` by yourself), when you try to require `com.walmartlabs.lacinia` the load fails with `CompilerException java.lang.ClassNotFoundException: clojure.set, compiling:(com/walmartlabs/lacinia/validation/no_unused_fragments.clj:36:31)`


wilkerlucio:
Sorry, just realized the code here is different from the latest here, this might be already fixed (since I can't see usages of `clojure.set` on this file anymore).

PR #31, opened 2017-04-05
Fill gaps in type support

hlship:
This fills the gaps in type support, allowing `(not-null (list (not-null Person)))` or `(list (list String))` or even more convoluted changes.

The role of the field resolver has simplified: previously, a field resolver would return a value and the library would wrap the field resolver to apply a number of validations.

Now the wrapping is minor and a new concept, a selector function, is responsible for validations, simple transformations (applying a tag automatically for concrete types, for example), and dealing with lists (the `mapv` has moved into the selector code).

Types are now structured more like they are in the introspection schema; a field definition will have a :type key whose values is a type map. The type map contains a :kind that can be :list, :non-null, or :root, and a nested :type.  For :list and :non-null, the nested type is another type map.  For :root it is a schema element name.

Previously, the type map was flat and just had keys :multiple? and :non-nullable?.

There's been some ripple effect changes into supporting arguments and variables and defaults on arguments and variables.

The changes to the introspection tests reflect that the proper wrapper types are being reported; previous test expectations were incorrect.

Benchmarking saw a small uptick on parse time, partially offset by a decrease in execution time.


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -228,49 +221,109 @@
...
+  ;; TODO: This nested maps format works, but given the simple modifiers
+  ;; we have, just converting from nested lists to a flattened vector
+  ;; might work just as well. It would also make finding the root type
+  ;; cheap: just use last.
+  [type]
+  (cond
+    (list? type)
+    (let [[modifier next-type & anything-else] type
+          kind (get {'list :list
+                     'non-null :non-null} modifier)]
```
hlship: Was thinking we could add `!` as a synonym for `non-null`, and `*` as a synonym for `list`, e.g., `:type (* (! String))`.

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -289,221 +342,156 @@
                                 [arg-name (compile-arg arg-name arg-def)])
                               %))))
 
-;;-------------------------------------------------------------------------------
```
hlship: All the enforcers logic has been transformed into the selectors.

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -289,221 +342,156 @@
...
+                                    (q field-name)
+                                    (-> object-def :type-name q)
+                                    (-> field-def :type q))
+                            {:field field-def
+                             :field-name field-name
+                             :schema-types (type-map schema)})))
+        category (:category field-type)
+
+        ;; Build up layers of checks and other logic.
+        coercion-wrapper (when (= :scalar category)
```
hlship: This is like the old enforcers code, where we use the information in the schema once, at compile time, to make decisions so that we don't have to keep making those decisions again and again during query executions.

src/com/walmartlabs/lacinia/internal_utils.clj
```diff
@@ -3,24 +3,6 @@
   {:no-doc true}
   (:require [clojure.string :as str]))
 
-(defn ensure-seq
```
hlship: I found some dead code.

src/com/walmartlabs/lacinia/introspection.clj
```diff
@@ -51,9 +51,9 @@
             ;; This is for resolve-nested-type, but it might actually be
             ;; a field definition, argument definition, or input value (a field of
             ;; an input object).
-            ::field-def arg-def})
```
hlship: Before ::field-def was actually a field def, a variable def, or an argument def.  Instead, we pass forward the type map from one of those.

src/com/walmartlabs/lacinia/introspection.clj
```diff
@@ -128,23 +128,25 @@
 (defn ^:private resolve-nested-type
   [context _ value]
   (let [schema (get context constants/schema-key)
-        field-def (::field-def value)
-        {:keys [type multiple? non-nullable?]} field-def]
-    (cond
-      multiple?
+        {:keys [kind type]} (::type-map value)]
+    (case kind
```
hlship: Here's where we end up modelling in the compiled schema that same kind of structure exposed in the introspection schema.

src/com/walmartlabs/lacinia/validation/no_unused_fragments.clj
```diff
@@ -1,4 +1,6 @@
-(ns com.walmartlabs.lacinia.validation.no-unused-fragments)
+(ns com.walmartlabs.lacinia.validation.no-unused-fragments
+  (:require
+    [clojure.set :as set]))
```
hlship: This was caught by the CI build, which does things in a different order than local builds for some reason.

test/com/walmartlabs/lacinia/enums_test.clj
```diff
@@ -1,28 +1,27 @@
...
     [com.walmartlabs.test-schema :refer [test-schema]]
     [com.walmartlabs.lacinia :refer [execute]]
-    [com.walmartlabs.lacinia.schema :as schema])
+    [com.walmartlabs.lacinia.schema :as schema]
+    [com.walmartlabs.test-reporting :refer [report]]
+    [com.walmartlabs.test-utils :refer [simplify]])
   (:import
     (clojure.lang ExceptionInfo)))
 
-(def ^:dynamic *compiled-schema* nil)
```
hlship: I stripped out the use of fixtures; since the schema is compiled and then immutable there's no need to discard and recreate it so often, and it's much easier to diagnose failed tests which the schema available.  I'm not sure about Cider, but in Cursive, running all or just one test will still do a reload of the test namespace (and any changed namespaces).

test/com/walmartlabs/lacinia/enums_test.clj
```diff
@@ -1,28 +1,27 @@
...
 
 (deftest can-provide-enum-as-bare-name
-  (is (= "Luke Skywalker"
-         (-> "{ hero(episode: NEWHOPE) { name }}"
-             q
-             :data
-             :hero :name))))
+  (let [response (q "{ hero(episode: NEWHOPE) { name }}")
+        hero-name (-> response :data :hero :name)]
+    (report response
```
hlship: Early on I was having a lot of trouble, and copied the eReceipts report macro in; once that's in place,. there isn't a real issue to leaving it around.

test/com/walmartlabs/lacinia_test.clj
```diff
@@ -1,21 +1,23 @@
...
-(use-fixtures :once
-              (fn [f]
-                (binding [*schema* (schema/compile test-schema)]
-                  (f))))
+(defn execute
+  "Executes the query but reduces ordered maps to normal maps, which makes
+  comparisons easier.  Other tests exist to ensure that order is maintained."
+  [schema q vars context]
+  (-> (lacinia/execute schema q vars context)
+      simplify))
```
hlship: `simplify` makes sure everything is a simple map or vector (no lists, no ordered maps). Much easier to compare diffs when tests fail that way.

candid82:
@hlship is this a breaking change? If so, can you write migration notes describing what may break and how to fix it?

hlship:
Not a breaking change, outside of one thing: Previously `(not-null (list X))` and `(list (not-null X))` were treated identically, which was invalid, and not spec-compliant. This just opens up more sophisticated type expressions than what was supported previously.

hlship:
Closing until I get this merged.

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -209,51 +229,87 @@
...
+                                                 schema/type-tag)
+                          field-name (:field selection)]
+                      (-> execution-context
+                          :context
+                          (get constants/schema-key)
+                          (get concrete-type-name)
+                          :fields
+                          (get field-name)
+                          :selector
+                          (or (throw (ex-info "Sanity check: no selector."
```
hlship: It did while I was debugging, and I've left sanity checks in place before.

hlship:
Yes, bigger than I hoped BUT we're talking about fixing something pretty fundamental to GraphQL: the type system.

PR #35, opened 2017-04-06
Remove unused bindings / namespaces

candid82:
Removing a few unused bindings / namespaces Joker found.

### DIFF
--- a/test/com/walmartlabs/test_utils.clj
+++ b/test/com/walmartlabs/test_utils.clj
@@ -2,7 +2,10 @@
   (:require
     [clojure.test :refer [is]]
     [clojure.spec.test :as stest]
-    [clojure.walk :as walk])
+    [clojure.walk :as walk]
+
+    ;; Needed for side effects (defines OrderedMap), do NOT remove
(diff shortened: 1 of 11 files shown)


test/com/walmartlabs/test_utils.clj
```diff
@@ -2,7 +2,6 @@
   (:require
     [clojure.test :refer [is]]
     [clojure.spec.test :as stest]
-    flatland.ordered.map
```
hlship: This one should be left; it ensures that the OrderedMap class exists for the :import.

bcarrell: I don't think that's true.

hlship: The `deftype` for `OrderedMap` is in that namespace, so if you want to dependably load `OrderedMap`, the namespace should exist first. `OrderedMap` is not in the flatland jar.

candid82: Ah, good catch! I put it back in.

Issue #34, opened 2017-04-05
Custom scalar :parse is never called


ura1a:
When defining a custom scalar that will be sent as an argument, the `:parse` function is never called, but `:serialize` is. Example below will return with an error, but wouldn't if the parsing had been called:

```
(require
  '[clojure.spec :as s]
  '[com.walmartlabs.lacinia :as g]
  '[com.walmartlabs.lacinia.schema :as schema])

(import
  java.text.SimpleDateFormat
  java.util.Date)

(def date-formatter
  "Used by custom scalar :Date."
  (SimpleDateFormat. "yyyy-MM-dd"))

(defn parse-fn [input]
  (prn "PARSE NEVER CALLED!")
  (.parse date-formatter input))

(defn serialize-fn [output]
  (prn "SERIALIZE CALLED AND WILL THROW BECAUSE NEVER PARSED!")
  (.format date-formatter output))

(def schema
  (schema/compile
    {:scalars {
               :Date {:parse (s/conformer parse-fn)
                      :serialize (s/conformer serialize-fn)}}

     :queries {
               :today {:type :Date
                       :args {:asOf {:type :Date}}
                       :resolve (fn [ctx args v] (:asOf args))}}}))

(g/execute schema "query($asOf: Date!){today(asOf: $asOf)}" {:asOf "2017-04-05"} nil)
```

annapawlicka:
I'm looking into this

hlship:
Which version of the library does this affect?

PR #36, opened 2017-04-06
Apply custom scalar's parse fn to resolved variables 

annapawlicka:
This PR fixes https://github.com/walmartlabs/lacinia/issues/34

Variables can be custom scalars, and so we should treat them as such when their value is resolved.

### DIFF
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -500,7 +500,7 @@
       (process-literal-argument schema {:type argument-definition} [:scalar result])
 
       :else
-      nil)))
+      result)))
 
 (defmethod process-dynamic-argument :variable
   [schema argument-definition [_ arg-value]]
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -465,11 +465,42 @@
   (fn [schema argument-definition [arg-type arg-value]]
     arg-type))
 
-
 (defn ^:private non-null-kind?
   "Peeks at the kind of the provided def (field, argument, or variable) to see if it is :non-null"
   [any-def]
   (-> any-def :type :kind (= :non-null)))
+
+(defn ^:private process-result
+  "Checks result against variable kind, iterates over nested types, and applies respective
+  actions, if necessary, e.g. parse for custom scalars."
+  [schema result argument-definition arg-value]
+  (let [next-type (:type argument-definition)
+        scalar-type? (and (= :root (:kind argument-definition))
+                          (scalar? (get schema next-type)))
+        list-type? (= :list (:kind argument-definition))
+        non-null-kind? (= :non-null (:kind argument-definition))]
+    (cond
+      ;; we can only hit this if we iterate over list members
+      (and (nil? result) non-null-kind?)
+      (throw-exception (format "Variable %s contains null members but supplies the value for a list that can't have any null members."
+                               (q arg-value))
+                       {:variable-name arg-value})
+
+      list-type?
+      (if (not (sequential? result))
+        (throw-exception (format "Variable %s doesn't contain the correct number of (nested) lists."
+                                 (q arg-value))
+                         {:variable-name arg-value})
+        (mapv #(process-result schema % next-type arg-value) result))
+
+      (and (some? result) (map? next-type))
+      (process-result schema result next-type arg-value)
+
+      (and (some? result) scalar-type?)
+      (process-literal-argument schema {:type argument-definition} [:scalar result])
+
+      :else
+      nil)))
 
 (defmethod process-dynamic-argument :variable
   [schema argument-definition [_ arg-value]]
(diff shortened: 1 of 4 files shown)


test/com/walmartlabs/lacinia/custom_scalars_test.clj
```diff
@@ -0,0 +1,151 @@
...
+(deftest custom-scalar-query
+  (let [q "{ now { date }}"]
+    (is (= {:data {:now {:date "A long time ago"}}}
+           (execute default-schema q nil nil)))))
+
+(deftest custom-scalars
+  (testing "custom scalars defined as conformers"
+    (let [parse-conformer (s/conformer
+                           (fn [x]
+                             (prn "qoooooooo")
```
bcarrell: Left a `prn` in there.

kgann: 👀

annapawlicka: Ooops haha

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -491,34 +491,39 @@
     (let [{:keys [default-value]} argument-definition
           non-nullable? (non-null-kind? argument-definition)
           var-non-nullable? (non-null-kind? variable-def)
-          var-default-value (:default-value variable-def)]
+          var-default-value (:default-value variable-def)
+          variable-type (get-in argument-definition [:type :type])]
```
hlship: This assumes that the argument definition's type is `{:kind :root :type scalar-name}` and doesn't take into account simple things like a `:non-null` kind wrapper or (more involved) a `:list` kind wrapper.

test/com/walmartlabs/lacinia/custom_scalars_test.clj
```diff
@@ -0,0 +1,150 @@
...
+                                         :line 1}]
+                            :message "Invalid value for a scalar type."
+                            :query-path [:events
+                                         :lookup]
+                            :type :Event
+                            :value "1"}]}
+                 (execute schema q2 nil nil))
+              "should return error message"))))))
+
+(deftest custom-scalars-with-variables
```
hlship: An additional test or two, for `(not-null scalar-type)` and `(list scalar-type)` are needed.  I believe the code as currently implemented will not handle these cases.

test/com/walmartlabs/lacinia_test.clj
```diff
@@ -686,87 +681,3 @@
                  :default-value "0001"}}}}}]
     (is-thrown [e (schema/compile schema-non-nullable-with-defaults)]
       (is (= (.getMessage e) "Field `id' of type `person' is both non-nullable and has a default value.")))))
-
-(deftest custom-scalars
```
hlship: I'm very glad to see lacinia-test start to break up into smaller bits!

src/com/walmartlabs/lacinia/parser.clj
bcarrell: Good call, still grokking the new type changes.

annapawlicka: I didn't understand new latest changes to types, but I do now. Fun times:
```
{:type {:kind :non-null, :type {:kind :list, :type {:kind :non-null, :type {:kind :root, :type :Date}}}}, :arg-name :between}
```

test/com/walmartlabs/lacinia/custom_scalars_test.clj
annapawlicka: I added a bunch of tests with a combination of not-nulls and lists.

annapawlicka:
@hlship @bcarrell I think I covered all cases, but let me know if there's something that I missed. The solution isn't as nice as I'd like it to be, but it's easiest for me to wrap my head around in this complex world of nested types and kinds.. I may give it another go once this new knowledge settles in my head.

bcarrell:
Looks like this should work, to me.  I'll let @hlship merge if he's happy.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -465,12 +465,43 @@
...
-
 (defn ^:private non-null-kind?
   "Peeks at the kind of the provided def (field, argument, or variable) to see if it is :non-null"
   [any-def]
   (-> any-def :type :kind (= :non-null)))
 
+(defn ^:private process-result
+  "Checks result against variable kind, iterates over nested types, and applies respective
+  actions, if necessary, e.g. parse for custom scalars."
+  [schema result argument-definition arg-value]
```
hlship: An argument definition contains the argument type; I think (I'll check) that you are passing an argument-type here (the :type key of an argument definition).

hlship: Yes, on line 529.  So I might call this `argument-type`, and rename `next-type` to `nested-type` or `wrapped-type`.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -465,12 +465,43 @@
...
+
+      list-type?
+      (if (not (sequential? result))
+        (throw-exception (format "Variable %s doesn't contain the correct number of (nested) lists."
+                                 (q arg-value))
+                         {:variable-name arg-value})
+        (mapv #(process-result schema % next-type arg-value) result))
+
+      (and (some? result) (map? next-type))
+      (process-result schema result next-type arg-value)
```
hlship: How about `recur`?

annapawlicka:
@hlship Done refactoring, back to you.

PR #37, opened 2017-04-07
Support field resolvers that return an promise, rather than a direct value

hlship:
This implements the discussion we've had about resolvers; the terminology has changed slightly to reflect that the returned values are Promises.

With one exception, no tests have changed (new tests were added).

Performance is just about the same as before.


hlship:
I'll go tweak the tests so they run on the CI.

test/com/walmartlabs/lacinia/directives_test.clj
```diff
@@ -208,7 +208,7 @@
                  id
                }
              }"]
-      (is (= {:data nil}
```
hlship: I believe this change is harmless. I'm not sure which of these actually matches the scenario best.

hlship:
I'll take care of the conflicts and re-open it shortly.

hlship:
Ready for review.

hlship:
I've been thinking about trying to reproduce all the Javascript reference implementation tests as a way to verify things like null collapse and other less obvious behaviors.

src/com/walmartlabs/lacinia/resolve.clj
```diff
@@ -4,32 +4,86 @@
...
   ([resolved-value resolver-errors]
    (->ResolverResultImpl resolved-value resolver-errors)))
+
+(defn resolve-promise
+  "Returns a ResolverResultPromise.
+
+   A value must be resolved and ultimately provided via [[deliver!]]."
+  []
+  (let [realized-result (promise)
+        callback-promise (promise)]
```
hlship: Because it has only two states: initial state, and final state.

Also, an earlier version of the code relied on the blocking-until-realized behavior of a promise; that's gone (along with the accessor methods) but there was no reason to change the realized-result or callback-promise to an atom, again, because of the exactly-two-states issue.

hlship: Effectively, the promise is a subclass of the normal one. The ResolverResult is simply an immediately realized version of the promise. This simplifies everything inside c.w.l.executor, as it can act on the ResolverResult exactly as if it was a promise, and it will work even if it is the simple (pre-realized) result.

hlship:
And so .... ?

PR #39, opened 2017-04-12
Simplify benchmark reporting

hlship:
I've simplified the code inside perf.clj to just collect and store the benchmarks; this gets rid of Incanter as a dev dependency.

I'll follow up with R scripts to generate the desired graphs from the CSV file, once I have that figured out.

### DIFF
--- a/dev-resources/perf.clj
+++ b/dev-resources/perf.clj
@@ -1,9 +1,6 @@
 (ns perf
   "A namespace where we track relative performance of query parsing and execution."
   (:require
-    [incanter.core :refer [with-data save conj-rows]]
-    [incanter.io :refer [read-dataset]]
-    [incanter.charts :refer [line-chart]]
     [org.example.schema :refer [star-wars-schema]]
     [criterium.core :as c]
     [com.walmartlabs.lacinia :refer [execute execute-parsed-query]]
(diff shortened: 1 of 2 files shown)


dev-resources/perf.clj
```diff
@@ -259,21 +237,38 @@
...
+            io/reader
+            csv/read-csv)
+        parse-double (fn [row ix]
+                       (update row ix #(Double/parseDouble %)))]
+    (into [header-row]
+          (->> raw-data
+               (mapv #(parse-double % 3))
+               (mapv #(parse-double % 4))))))
+
 (defn run-benchmarks [options]
```
annapawlicka: Both `run-benchmarks` and `read-dataset` could be made private since they're not used outside of this namespace.

project.clj
```diff
@@ -10,7 +10,8 @@
                  [org.flatland/ordered "1.5.4"
                   :exclusions [org.clojure/tools.macro]]]
   :profiles {:dev {:dependencies [[criterium "0.4.4"]
-                                  [incanter "1.5.7"]
+                                  [clj-time "0.13.0"]
```
annapawlicka: I don't see `clj-time` being used anywhere

hlship: Really, just need joda-time, to support the scalar tests.

candid82:
@hlship what's the reason for moving from incanter to R?

hlship:
Network effect.  Lack of documentation and support for Incanter vs. lots of that for R.  I just need to make a couple of simple charts from the benchmarks.csv file.

hlship:
I'd love to see this merged; I'd also like to do a release of Lacinia today.

dev-resources/perf.clj
```diff
@@ -259,21 +237,41 @@
...
+  []
+  (let [[header-row & raw-data]
+        (-> (io/file dataset-file)
+            io/reader
+            csv/read-csv)
+        ;; read-csv reads all columns as strings, so we need to do some
+        ;; work to get them back to numbers.
+        parse-double (fn [row ix]
+                       (update row ix #(Double/parseDouble %)))]
+    (into [header-row]
```
bcarrell: No reason not to use the transducer arity here, but it's super minor.

project.clj
```diff
@@ -10,7 +10,8 @@
                  [org.flatland/ordered "1.5.4"
                   :exclusions [org.clojure/tools.macro]]]
   :profiles {:dev {:dependencies [[criterium "0.4.4"]
-                                  [incanter "1.5.7"]
+                                  [joda-time "2.9.7"]
```
bcarrell: We should just use `java.time`, but I'll allow it.

PR #40, opened 2017-04-13
Optimization to resolvers

hlship:
This was a tiny optimization that had a larger impact than I expected.

Currently, whenever a field resolver is invoked, we have to do a check to see if a normal value, or a ResolverResult, is returned.

This change makes that optional: a function whose meta :tag is ResolverResult can skip that wrapping step and the `satisifies?` call.

The default-field-resolver factory was changed to return a function with the tag; that's where the savings is coming from, all the automatic fields get this behavior automatically, and that's the majority of fields in any query.

This can be merged after https://github.com/walmartlabs/lacinia/pull/39.


hlship:
I'm getting tempted to merge this myself.

candid82:
lgtm, if it really makes a difference performance wise. One of the things that I liked about the original design (resolvers always return tuples) is that we would not have to implement this "user assisted" optimization.

hlship:
Time to merge?

PR #41, opened 2017-04-18
Error maps may include parsed arguments, causing failures if those parsed values are not JSON serializable

kgann:
This PR will prevent returning conformed scalar arguments when resolvers return an error.


hlship:
LGtM

hlship:
We need another reviewer ...

PR #43, opened 2017-04-20
Support specifying operation-name in execute

brendanyounger:
Adds an optional parameter to `com.walmartlabs.lacinia/execute` which allows the caller to specify an operation name to execute for a given query/mutation.


src/com/walmartlabs/lacinia.clj
```diff
@@ -68,10 +68,10 @@
   : Additional data that will ultimately be passed to resolver functions.
 
   This function parses the query and invokes [[execute-parsed-query]]."
-  [schema query variables context]
+  [schema query variables context & [operation-name]]
```
hlship: I think I would prefer to see an optional map of additional options, :operation-name being one of them.  This makes it more reasonable to add further options in the future, and to pass the options does to parse-query and execute, as needed.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -739,9 +739,9 @@
 
     (and single-op?
          operation-name
-         (or (not (< 2 (count (second first-op))))
+         (or (not (< 2 (count first-op)))
```
hlship: Is this a bug in the existing code, or something else entirely?

test/com/walmartlabs/lacinia_test.clj
```diff
@@ -54,6 +54,19 @@
...
+  (let [q "query heroNameQuery { hero { id name } } query dummyQuery { hero { id } }"]
+    (is (= {:data {:hero {:id "2001" :name "R2-D2"}}}
+           (execute default-schema q {} nil "heroNameQuery"))))
+
+  (let [q "mutation changeHeroNameMutation ($from : String, $to: String) { changeHeroName(from: $from, to: $to) { name } }
+           query dummyQuery { hero { id } }"]
+    (is (= {:data {:changeHeroName {:name "Solo"}}}
+           (execute default-schema q {:from "Han Solo"
+                                      :to "Solo"}
+                    nil "changeHeroNameMutation")))))
```
hlship: Do we have checks and tests for an invalid query name being passed into execute?

src/com/walmartlabs/lacinia.clj
brendanyounger: Are you in favor of `& {:keys [operation-name]}` or `& [{:keys [operation-name]}]`?

src/com/walmartlabs/lacinia/parser.clj
brendanyounger: Seems to be a bug.  Original code was trying to call `(nth 2 %)` on a keyword (IIRC).

test/com/walmartlabs/lacinia_test.clj
brendanyounger: Will add.

src/com/walmartlabs/lacinia/parser.clj
hlship: That's bad, that code obviously hasn't been properly tested since the switchover from Instaparse to Antlr.

src/com/walmartlabs/lacinia.clj
hlship: A 4 arity and a 5 arity; the 5th argument is a map.  pseudo-keyword args are for things you use frequently at the REPL and get in the way of all other kinds of composition.

brendanyounger:
Implemented the proposed changes to arguments and added an additional test.

src/com/walmartlabs/lacinia.clj
```diff
@@ -68,10 +68,10 @@
   : Additional data that will ultimately be passed to resolver functions.
 
   This function parses the query and invokes [[execute-parsed-query]]."
-  [schema query variables context]
+  [schema query variables context & [{:keys [operation-name] :as opts}]]
```
kgann: > A 4 arity and a 5 arity; the 5th argument is a map.

I believe @hlship was requesting something like:

```clojure
([schema query variables context]
 (execute schema query variables context {})
([schema query variables context options]
 ...)
```

src/com/walmartlabs/lacinia.clj
```diff
@@ -36,12 +36,12 @@
     :else
     (try
       (executor/execute-query (assoc context
-                                     constants/schema-key schema
-                                     constants/parsed-query-key prepared))
+                                constants/schema-key schema
```
hlship: We really don't want to get into an indentation war over this code.

hlship:
At some future date we may juggle things such that we select the operation during the prepare, not parse, stage.  That may be an incompatible change, which is why Lacinia is still alpha.

brendanyounger:
Sorry about the indentation changes.  I use Cursive's auto-indent without even thinking about it.

hlship:
I use Cursive as well, but most of the team uses Emacs, and that is the de facto standard on indentation.  So just be careful to only indent in code you've changed (that's an option somewhere in Cursive too, I think).

hlship:
I'd like to see another review and merge this.  Right now it is in limbo.

bcarrell:
Merged this, thank you!

PR #47, opened 2017-04-23
Fix issue w/ vars and enums, make sure benchmark queries work, update documentation

hlship:
This started as just updating the documentation, and along the way, I hit a bug with variables assigned to enum fields, which I've fixed.


src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -491,33 +491,43 @@
...
+    :let [nested-type (:type argument-type)
+          kind (:kind argument-type)]
 
-      list-type?
-      (if (not (sequential? result))
-        (throw-exception (format "Variable %s doesn't contain the correct number of (nested) lists."
-                                 (q arg-value))
-                         {:variable-name arg-value})
-        (mapv #(process-result schema % nested-type arg-value) result))
+         ;; we can only hit this if we iterate over list members
```
bcarrell: It seems like indentation gets screwed up from here down.

hlship: Looks like Cursive "forgot" how to indent cond-let.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -491,33 +491,43 @@
...
+    :let [nested-type (:type argument-type)
+          kind (:kind argument-type)]
+
+    ;; we can only hit this if we iterate over list members
+    (and (nil? result) (= :non-null kind))
+    (throw-exception (format "Variable %s contains null members but supplies the value for a list that can't have any null members."
+                             (q arg-value))
+                     {:variable-name arg-value})
+
+    (= :list (:kind argument-type))
```
annapawlicka: `(= :list kind)`

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -491,33 +491,43 @@
...
-      :else
-      result)))
+    (= category :scalar)
+    (process-literal-argument schema {:type argument-type} [:scalar result])
+
+    ;; enums have to be handled carefully because they are likely strings in
+    ;; the variable map.
+
+    (= category :enum)
+    (process-literal-argument schema {:type argument-type} [:enum (as-keyword result)])
```
annapawlicka: Could we add a test for this scenario so that we avoid regression in the future?

hlship:
Ok, back in your court.

PR #49, opened 2017-04-26
Add function c.w.l/execute-parsed-query-async

hlship:
Adds a new function, `execute-parsed-query-async` that returns a ResolverResult.  This is a good bridge for pedestal-lacinia to use in order to provide an async handler for executing queries.

### DIFF
--- a/src/com/walmartlabs/lacinia.clj
+++ b/src/com/walmartlabs/lacinia.clj
@@ -3,7 +3,8 @@
             [com.walmartlabs.lacinia.constants :as constants]
             [com.walmartlabs.lacinia.executor :as executor]
             [com.walmartlabs.lacinia.validator :as validator]
-            [com.walmartlabs.lacinia.internal-utils :refer [cond-let to-message]])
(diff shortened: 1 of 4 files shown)


src/com/walmartlabs/lacinia.clj
```diff
@@ -3,17 +3,21 @@
...
   [(merge {:message (to-message exception)}
           (ex-data exception))])
 
-(defn execute-parsed-query
+(defn execute-parsed-query-async
   "Prepares a query, by applying query variables to it, resulting in a prepared
-  query which is then executed."
+  query which is then executed.
+
+  Returns a [[ResolverResult]] around the response value (with :data and/or :errors keys)."
```
bcarrell: Same question as the other PR, is the `[[...]]` for Codox?  Are we even using that?

hlship: Yes, the `[[..]]` syntax creates a hyperlink to the referenced thing. It's pretty smart about finding it, even in other namespaces.

PR #50, opened 2017-04-26
Collect timing information from resolvers (optionally)

hlship:
This builds on my prior PR, #49, which should be merged first.

This tracks time to resolve values in non-default resolver functions, and exposes it in the :extensions key of the response map.

I'm pretty happy with how this came out; I think that structuring the data in terms of the query tree, rather than the schema, is the right way to go.

*Update:* I'm starting to think that the query tree is not the right way to go, as it effectively prevents you from aggregating timing data between different queries; if the timings were organized instead by object and field, it would be simple to aggregate. Thoughts?

### DIFF
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -130,14 +130,10 @@
                                        timing {:start start-ms
                                                :finish finish-ms
                                                ;; This is just a convienience:
-                                               :elapsed elapsed-ms}]
-                                   ;; The extra key is to handle a case where we time, say, [:hero] and [:hero :friends]
-                                   ;; That will leave :friends as one child of :hero, and :execution/timings as another.
-                                   ;; The timings are always a list; we don't know if the field is resolved once,
-                                   ;; resolved multiple times because it is inside a nested value, or resolved multiple
-                                   ;; times because of multiple top-level operations.
+                                               :elapsed elapsed-ms}
+                                       {:keys [field-name containing-type-name]} (:field-definition field-selection)]
                                    (swap! timings
-                                          update-in (conj (:query-path field-selection) :execution/timings)
+                                          update-in [containing-type-name field-name]
                                           (fnil conj []) timing)))
 
                                (when-let [errors (-> resolve-errors
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -89,13 +89,6 @@
                          :value value
                          :value-meta (meta value)})))))
 
-(defmacro ^:private now
-  []
-  `(System/currentTimeMillis))
-
-(def ^:private conj-nil
-  (fnil conj []))
-
 (defn ^:private invoke-resolver-for-field
   "Resolves the value for a field selection node, by passing the value to the
   appropriate resolver, and passing it through a chain of value enforcers.
@@ -121,7 +114,7 @@
             field-resolver (field-selection-resolver schema field-selection container-value)
             start-ms (when (and (some? timings)
                                 (not (-> field-resolver meta ::schema/default-resolver)))
-                       (now))
+                       (System/currentTimeMillis))
             resolver-result (try
                               (field-resolver resolve-context arguments container-value)
                               (catch Throwable t
@@ -132,7 +125,7 @@
         (resolve/on-deliver! resolver-result
                              (fn [resolved-value resolve-errors]
                                (when start-ms
-                                 (let [finish-ms (now)
+                                 (let [finish-ms (System/currentTimeMillis)
                                        elapsed-ms (- finish-ms start-ms)
                                        timing {:start start-ms
                                                :finish finish-ms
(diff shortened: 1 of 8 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -75,7 +75,7 @@
   "Creates a clojure.spec/conformer as a wrapper around the supplied function.
 
   The function is only invoked if the value to be conformed is non-nil.
-
+]
```
thorstadt: Is this `]` intentional?

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -89,6 +89,12 @@
                          :value value
                          :value-meta (meta value)})))))
 
+(defmacro ^:private now
```
hlship: I think I had a master plan, but can't remember what it is.

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -117,6 +131,22 @@
...
+                                               ;; This is just a convienience:
+                                               :elapsed elapsed-ms}]
+                                   ;; The extra key is to handle a case where we time, say, [:hero] and [:hero :friends]
+                                   ;; That will leave :friends as one child of :hero, and :execution/timings as another.
+                                   ;; The timings are always a list; we don't know if the field is resolved once,
+                                   ;; resolved multiple times because it is inside a nested value, or resolved multiple
+                                   ;; times because of multiple top-level operations.
+                                   (swap! timings
+                                          update-in (conj (:query-path field-selection) :execution/timings)
+                                          conj-nil timing)))
```
hlship: True; we don't care so much about performance when we know we're taking a hit just to record the timings.

src/com/walmartlabs/lacinia/schema.clj
hlship: You aren't aware of the "leading ] in a docstring" convention? See, this is what happens when you don't go to the Clojure meetups.

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -98,16 +97,24 @@
   The final ResolverResult will always contain just a resolved value; errors are handled here,
   and not passed back in the returned ResolverResult.
 
+  Optinally updates the timings inside the execution-context with start/finish/elapsed time
```
thorstadt: Should be "optionally"

candid82:
@hlship needs master merge for clean diff

hlship:
Ok, one sec.

candid82:
I think I like the original structure of the timings data (query tree) better than the new one  (schema tree). It makes it easier to see where the time is spent for a particular query, which is in my mind the main use case for this feature. Not sure if it makes a lot of sense to aggregate timings across different queries on per-resolver basis. I'd also suggest making `start` and `finish` timestamps optional as they clutter the output and aren't very useful most of the time. Have you thought of tracking elapsed time in nanoseconds instead of milliseconds?

hlship:
The problem with the query tree approach is that it's very specific to the query, which may be ephemeral ... but can only reference what's in the schema.

Can you elaborate on what you could identify using the query structure approach that you can't using the schema structure approach?

I'm not too worried about clutter, a filter can easily remove those keys, or collapse to just :elapsed and probably that should be kept on the server and aggregated anyway.  I'm kind of envisioning a system where periodically, a single request though gets timed and the result aggregated, but the timing data doesn't get sent to the client.

I considered nanoseconds, but the extra accuracy isn't useful for real delays (e.g., getting data from Cassandra) and would just be noise for trivial resolvers, such as the default resolvers.

candid82:
The use case I had in mind was different: you (as a client) send a graphql query to the server and get timing data as part of the response, and visually examine it to get a quick idea where the time is spent and how adding / removing fields affects timing. It's possible with schema structure approach, just less convenient. The fact that the timing data is very specific to the query is a good thing for my use case, not a problem.

bcarrell:
On the points above, I feel it depends who we're optimizing for.  If it's the user of the service, then I think having timing data in the shape of the query itself is ideal because they can match everything 1:1 (here's the data, here's the timing of everything in that shape, can I adjust my query to remove things I don't particularly need and bring down the total time?).  If it's the maintainers/developers of the service, then having it in the shape of the schema is better (this resolver on this field is slow, let's go optimize it).

I think I'm on board with @candid82's points here.  I would take some inconvenience on my side so that the users of the service understand what their queries are doing a little easier.

Maybe it's possible to achieve both goals?  Control the output shape using a parameter instead of a boolean?

hlship:
I disagree, I think this format is less clear for both client and server-side developers.  I've reverted the change so you can take a proper look at it.

bcarrell:
Can you explain why you feel the query format is less clear to a client than the schema format?  When I'm using the API, I often think in terms of "query paths" and not schema definitions.  When I'm making queries it's about the field names and not the names of the types, to me.

hlship:
I think it's less obvious with the tiny Star Wars schema, but on a real schema, with more levels, the query-oriented schema map will be quite sparse, leading to a lot of empty nodes. I don't like the easy naming conflicts (the 'human' query vs. the 'human' object).  There may be some value to the nesting ... e.g., the `gnip` resolver uses a lot of time, but only inside the `fnord` query, not the `quulx` query. 

I don't think that making it configurable is a good option; that feels like a cop-out.  I'd rather release it as-is, with a note that the format is still being evaluated, may change, feedback wanted.

bcarrell:
I'll defer to you: if you believe the schema representation is better, let's go with that.  Regardless, I think it should be noted, as you said, that the format is being evaluated and that feedback is requested.

PR #54, opened 2017-05-02
Add ability to control error message from a scalar transformer coercion

hlship:
This gives users the option  to  customize the error message when a scalar can't be transformed.  This will be quite valuable for some of the improvements to schemas we've been discussing.


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -14,7 +14,7 @@
     [com.walmartlabs.lacinia.internal-utils
      :refer [map-vals map-kvs filter-vals deep-merge q
              is-internal-type-name? sequential-or-set? as-keyword
-             combine-results]]
+             combine-results cond-let]]
```
annapawlicka: `cond-let` isn't used anywhere in this ns

src/com/walmartlabs/lacinia/util.clj
```diff
@@ -25,25 +27,34 @@
            :else
            ;; If resolver-k is not a keyword, it must be a sequence,
            ;; in which first element is a key that points to a resolver
-           ;; factory in resolver-m and subsequent elements are arguments
+           ;; factory in resolver-m and subsequent elements are argumentsc
```
annapawlicka: Keybinding gone wrong?

src/com/walmartlabs/lacinia/schema.clj
hlship: Well, maybe it should be :-)

bcarrell:
Conflicts

hlship:
Wait, I'm responsible for conflicts while I sleep?  

I'll fix this up when I get to the office.

hlship:
Should be ready to go.

PR #57, opened 2017-05-08
Make the default field resolver simpler

hlship:
Some of the top Lacinia feedback has been concerns about the default field resolver; this PR changes the behavior to be the simplest possible (no translation of underscores to dashes), but leaves the old behavior easily accessible via a schema/compile option.

This also simplifies documentation.


project.clj
```diff
@@ -10,6 +10,7 @@
                  [org.flatland/ordered "1.5.4"
                   :exclusions [org.clojure/tools.macro]]]
   :profiles {:dev {:dependencies [[criterium "0.4.4"]
+                                  [org.clojure/test.check "0.9.0"]
```
hlship: I'm mystified as to why this was suddenly necessary.

PR #59, opened 2017-05-09
Don't use metadata to identify an object's type

### LINKED ISSUE #55: Exception thrown if a field resolver returns anything but a IObj
If a lacinia resolver returns anything other than `clojure.lang.IObj` an exception is being generated. The reason for this is that it is somewhere in the call chain trying to attach meta-data to the returned value.

`java.lang.ClassCastException: datomic.query.EntityMap cannot be cast to clojure.lang.IObj`

This however is a bit awkward when working with plain Java return types like e.g. `EntityMap` from datomic. 

We have worked around this by defining a custom resolver function (needed anyway to do namespaced keywords conversion).

```clojure
(defn find-object-type
  [v]
  (some->> v
           (keys)
           (filter #(= "id" (name %)))
           (first)
           (namespace)))

(defn default-field-resolver
  "This is a function that accepts a field name (a keyword) and
  returns a field resolver function for the field. This includes
  transforming to datomic namespaced keywords and wrapping/unwrapping
  DatomicEntity objects."
  [field-name]
  ^ResolverResult (fn [ctx args v]
                    (let [value (if (instance? DatomicEntity v) ;; DatomicEntity is a record with only one field, the entity
                                  (get (.entity v) (keyword (find-object-type (.entity v)) (name field-name)))
                                  (get v field-name))]
                      (resolve-as
                        (if (instance? datomic.Entity value)
                          (wrappers/->DatomicEntity value)
                          value)))))
```

This is however only part of the solution as each root query needs to wrap it's datomic entity results in the clojure record.

We also would like our resolvers to return manifold deferred's instead of a lacinia promise because of code readability. This approach suffers the same issue as with a Datomic Entity and combining the two becomes more and more a problem.

Is there any infrastructure available to extend lacinia's default resolver functionality (not sure if I'm using the right terminology here). What would be sufficient for this use case:
1. a function that is applied before the returned resolver result is being passed down to the type tagging functionality. Here you could wrap everything.
2. the default resolver function that can unwrap results and get the values out for a given field (i.e. what you can specify now as ':default-field-resolver').

An example:

```clojure
(defn default-field-resolver
  [field-name]
  ^ResolverResult (fn [ctx args v]
                    (resolve-as
                      (if (instance? DatomicEntity v)
                        (get (.entity v) (keyword (find-object-type (.entity v)) (name field-name)))
                        (get v field-name)))))

(defn default-resolver-middleware
  [resolver]
  (fn [ctx args v]
    (let [result (resolver ctx args v)]
      (if (instance? datomic.Entity result)
        (wrappers/->DatomicEntity result)
        result))))

(lacinia.schema/compile
  {:default-field-resolver default-field-resolver
   :default-middleware     default-resolver-middleware})
```

hlship:
Fixes #55


src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -104,64 +100,63 @@
...
-                                 (swap! (:errors execution-context) into
-                                        (enhance-errors field-selection errors)))
-                               ;; That's it for handling errors, so just resolve the value and
-                               ;; not the errors.
-                               (resolve/deliver! final-result resolved-value)))
-        final-result)
-      ;; Else, not a field selection, but a fragment selection, which starts with the
-      ;; same resolved value as the containing field or selection.
-      (resolve/resolve-as container-value))))
+  (let [timings (:timings execution-context)
```
hlship: Previously, this could be invoked with a non-field-selection. That can't happen anymore, so this entire function is a bit simpler.

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -330,15 +322,14 @@
   Accumulates errors in the execution context as a side-effect."
   [execution-context selection]
   (let
-    [resolver-result (invoke-resolver-for-field execution-context selection)
+    [is-fragment? (-> selection :selection-type (not= :field))
```
hlship: Here's the start of changes to split the execution path for fields from fragments (inlined or named)

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -357,54 +350,57 @@
...
+      ;; Here's where it comes together.  The field's selector
+      ;; does the validations, and for list types, does the mapping.
+      ;; It also figures out the field type.
+      ;; Eventually, individual values will be passed to the callback, which can then turn around
+      ;; and recurse down a level.  The result is a map or a list of maps.
+
+      (let [final-result (resolve/resolve-promise)]
+        (resolve/on-deliver! (invoke-resolver-for-field execution-context selection)
+                             (fn [resolved-value _]
+                               (let [selector-result (selector resolved-value nil selector-callback)]
```
hlship: The selector signature changed, that nil is a placeholder for the field's concrete type (either supplied automatically, or extracted from a TaggedType).

CHANGES.md
```diff
@@ -5,6 +5,11 @@ it is now a direct mapping, without converting underscores to dashes.
 The old behavior is still available via an option to
 `com.walmartlabs.lacinia.schema/compile`.
 
+The function `com.walmartlabs.lacinia.schema/tag-with-type` has changed; it
+now returns an entirely different value (rather than the same value with
+different meta-data). This is to allow resolved values that are not maps
+(or sequences of maps).  The related function `com.walmartlabs.lacinia.schema/type-tag` has been removed.
```
candid82: More accurately, that do not support meta-data.

hlship:
Tweaked the changes, ready for merge.

PR #61, opened 2017-05-11
fix nested input arguments

### LINKED ISSUE #60: Error when providing nested input object via variables
This is a minimal case to reproduce (v 0.16.0)

**Schema**
```clojure
(def schema
  {:input-objects {:IDObject {:fields {:id {:type '(non-null String)}}}}
   :objects       {:Something  {:fields {:id         {:type '(non-null String)}
                                         :name       {:type 'String}
                                         :otherThing {:type :OtherThing}}}
                   :OtherThing {:fields {:id   {:type '(non-null String)}
                                         :name {:type 'String}}}}
   :mutations     {:CreateSomething
                   {:args {:name {:type 'String}
                           :otherThing {:type :IDObject}}
                    :type :Something}}})
```

**Execute the query by filling in parameters directly**
```clojure
(com.walmartlabs.lacinia/execute (com.walmartlabs.lacinia.schema/compile schema)
                                 "mutation CreateSomething { CreateSomething(name: \"Test\", otherThing: {id: \"ID\"}) {id }}"
                                 nil
                                 nil)
=> {:data #ordered/map([:CreateSomething nil])}
```

**Execute the query by providing variables**
```clojure
(com.walmartlabs.lacinia/execute (com.walmartlabs.lacinia.schema/compile schema)
                                 "mutation CreateSomething($name: String!, $otherThing: IDObject!) { CreateSomething(name: $name, otherThing: $otherThing) {id }}"
                                 {:name "Test"
                                  :otherThing {:id "ID"}}
                                 nil)
=> {:errors [{:message "Sanity check - no option in process-result."}]}
```

The error happens in `com.walmartlabs.lacinia.parser/process-result` because there is no matching clause in the cond for `(= category :input-object)`.

stijnopheide:
- create tests that cover the issue
- update parser process-result to recursively find the category of input objects

fixes #60

### DIFF
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -513,10 +513,10 @@
       (throw-exception (format "Variable %s doesn't contain the correct number of (nested) lists."
                                (q arg-value))
                        {:variable-name arg-value})
-      (mapv #(construct-literal-argument schema % nested-type arg-value) result))
+      [:array (mapv #(construct-literal-argument schema % nested-type arg-value) result)])
 
     (nil? result)
-    nil
+    [:null nil]
 
     (map? nested-type)
     (recur schema result nested-type arg-value)
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -496,7 +496,7 @@
   [any-def]
   (-> any-def :type :kind (= :non-null)))
 
-(defn ^:private determine-result-type
+(defn ^:private construct-literal-argument
   [schema result argument-type arg-value]
   (cond-let
     :let [nested-type (:type argument-type)
@@ -513,7 +513,7 @@
       (throw-exception (format "Variable %s doesn't contain the correct number of (nested) lists."
                                (q arg-value))
                        {:variable-name arg-value})
-      (mapv #(determine-result-type schema % nested-type arg-value) result))
+      (mapv #(construct-literal-argument schema % nested-type arg-value) result))
 
     (nil? result)
     nil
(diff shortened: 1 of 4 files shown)


src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -515,29 +513,43 @@
...
+    [:object (let [object-fields (get-in schema [nested-type :fields])]
+               (reduce (fn [acc k]
+                         (let [v (get result k)
+                               field-type (get object-fields k)]
+                           (assoc acc k (determine-result-type schema v field-type arg-value))))
+                       {}
+                       (keys result)))]
 
     :else
     (throw (IllegalStateException. "Sanity check - no option in process-result."))))
```
hlship: Should update this error message for the new function name.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -496,9 +496,7 @@
   [any-def]
   (-> any-def :type :kind (= :non-null)))
 
-(defn ^:private process-result
-  "Checks result against variable kind, iterates over nested types, and applies respective
-  actions, if necessary, e.g. parse for custom scalars."
+(defn ^:private determine-result-type
```
hlship: Not sure this name reflects what this function does.

First off `process-result` should be named something more like `substitute-variable`.  Sorry, it was like that when you got here, but first rule of boyscouts: leave things better than you found them.

`determine-result-type` is really creating a literal value from the variable, possibly recursively.  Perhaps 'construct-literal-argument` would be a better name?  Because that's what we're doing, constructing literal arguments (those tuples normally obtained via `xform-argument-value`).

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -515,29 +515,43 @@
...
-      (mapv #(process-result schema % nested-type arg-value) result))
+      (mapv #(determine-result-type schema % nested-type arg-value) result))
 
     (nil? result)
     nil
 
     (map? nested-type)
     (recur schema result nested-type arg-value)
 
-    :let [category (when (= :root kind)
```
stijnopheide: @hlship One question I had during this update is, why is this check for :root kind here? When we calculate the literal type recursively the check for :root doesn't work (nor make sense), but I'm not sure if this has any impact on root variable types. If there is an impact, the function needs to be rewritten to have a special case for the root element.

hlship: The old logic had gone through a few permutations; your corrected it. By the time we get to this clause we are on a root type.

PR #63, opened 2017-05-12
Add support for decoration of resolver functions

### LINKED ISSUE #58: Add ability to decorate resolver functions
A certain useful class of problems can be solved by passing field resolver functions through a set of middleware before being employed in a schema; this would allow for such things as authentication/authorization checks, implicit logging, tracing, and other common use-cases.

I would suggest a signature as follows:

     (middleware object-name field-name f) => f'

The object-name and field-name are keywords.  The function is as provided in the input schema, and the result is the same or a decorated resolver function.

hlship:
Fixes #58 

More work to document than to implement.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,11 @@ The old behavior is still available via an option to
 
 Please update your applications carefully.
 
+`compile` now has a new option, `:decorator`.
+The decorator is a callback applied to all non-default field resolvers.
+The primary use case is to adapt the return value of a field resolver,
+for example, from a core.async channel to a Lacinia ResolverResult.
+
 ## 0.16.0 -- 3 May 2017
 
 The function `com.walmartlabs.lacinia.schema/as-conformer` is now public.
--- a/dev-resources/com/walmartlabs/test_utils.clj
+++ b/dev-resources/com/walmartlabs/test_utils.clj
@@ -3,11 +3,9 @@
     [clojure.test :refer [is]]
     [clojure.spec.test :as stest]
     [clojure.walk :as walk]
-
-    ;; Needed for side effects (defines OrderedMap), do NOT remove
-    [flatland.ordered.map]
     [clojure.java.io :as io]
     [clojure.edn :as edn]
+    [com.walmartlabs.lacinia :as lacinia]
     [com.walmartlabs.lacinia.util :as util]
     [com.walmartlabs.lacinia.schema :as schema])
   (:import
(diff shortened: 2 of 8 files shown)


CHANGES.md
```diff
@@ -7,6 +7,11 @@ The old behavior is still available via an option to
 
 Please update your applications carefully.
 
+`compile` now has a new option, `:decorator`.
```
candid82: Following transducers tradition, I propose to name such decorators "transolvers" :)

dev-resources/com/walmartlabs/test_utils.clj
```diff
@@ -3,11 +3,9 @@
     [clojure.test :refer [is]]
     [clojure.spec.test :as stest]
     [clojure.walk :as walk]
-
-    ;; Needed for side effects (defines OrderedMap), do NOT remove
-    [flatland.ordered.map]
```
candid82: Not needed anymore?

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -523,8 +523,11 @@
   Adds a :selector function."
   [schema options containing-type field]
   (let [provided-resolver (:resolve field)
-        base-resolver (or provided-resolver
-                          ((:default-field-resolver options) (:field-name field)))
+        {:keys [default-field-resolver decorator]} options
+        field-name (:field-name field)
+        base-resolver (if provided-resolver
+                        (decorator (:type-name containing-type) field-name provided-resolver)
+                        (default-field-resolver field-name))
```
candid82: So decorators are not applied to default resolver? Should this be documented?

dev-resources/com/walmartlabs/test_utils.clj
hlship: No, because the `simplify` function now operates on any IPersistentMap.

CHANGES.md
hlship: Ever look at NetKernel with its "transreptors" (which can never be described, just experienced)?

src/com/walmartlabs/lacinia/schema.clj
hlship: This is documented.  Default resolvers are expected to be trivial, so they aren't included in timing and they are not subject to decoration.  If an application needs default resolvers to support the cross-cutting concerns addressed by decorator, they can provide their own default resolver generator as well.

hlship: Darn. It's only documented in the CHANGES.md.  Let me update.

hlship:
Improved the docs, ready for merge.

docs/_examples/decorator.edn
```diff
@@ -0,0 +1,32 @@
...
+        (let [resolve-promise (resolve/resolve-promise)]
+          (take! result
+            (fn [value]
+              (resolve/deliver! resolve-promise value nil)))
+          resolve-promise)))))
+
+
+(defn create-schema
+  []
+  (-> (io/resource "sample-scheam.edn")
```
sashton: Misspelled and missing file?

sashton:
@hlship What would a more complex decorator use case look like? 
For example: only a some resolvers return async channels, but not others, and only certain resolvers require extra authentication/privileges.
Do you have an idea of what a decorator would look like which handled such a scenario?

docs/_examples/decorator.edn
hlship: Yes, misspelled. Fixing.

Not a missing file, though it would be better if we extracted examples from actual executable code.

hlship:
More complicated cases: if I was building something more complex, I would lean on metadata on the function to describe if and who various decorators should be applied, much like how we use the `:type` tag to determine whether to apply the decorator to wrap the return value in a ResolverResult.

I purposely decided not to get involved with composing decorators. That could come later, if it is truly an issue.

hlship:
Ready for merge?

PR #67, opened 2017-05-15
Add an API for identifying operations in the query

### LINKED ISSUE #64: API for identifying operation(s) for request
Given a parsed query, it should be possible to identify the operations for the request.

Something like:

```
(query-operations parsed-query) => {:type :mutation :operations #{:add_employee :approve_vacation}}
```

This would be desirable as part of a web-tier pipeline, to enforce auth/auth before executing the query.

hlship:
Fixed #64

This is a quicky; it extracts the root fields and type of query (query or mutation) from the parsed query.

Perhaps in the future, there will be subscriptions as well, so it isn't just a boolean for mutation?.


sashton:
lgtm

PR #68, opened 2017-05-16
Allow field resolvers to identify nested fields in the selection tree

### LINKED ISSUE #65: API for identifying selections from application context
The application context passed to a field resolver function includes a key ::lacinia/selections, that identifies the field selections for the field.

We don't, and probably should not, publish the details of the structure of this value (as it is quite valuable to keep it private, opaque, and subject to change).

However, it exists so that field resolvers can preview the selections that may occur on the resolved value.

The canonical example of this is a resolver that performs a SQL database query, and may customize the request based on what is requested, for example, to optionally include a join only if the data from the join is needed.

I would suggest something like the following:

```
(selects-field? context :User/permissions) => true
```

This would scan the selections inside the context for any reference to the :permissions field of the :User object.

Another useful function might be:

```
(selections context) => #{:User/first_name :User/last_name :User:born_on :User/permissions :Permission/name}
```

I'm struggling with how this might work in the context of unions and interfaces; I would guess that the collection of these extended field names would driven by fragments, and ultimately, the fragments would identify the specific object fields.

hlship:
Fixes #65 

We've been doing this in our apps by pulling apart the selections context key, even though its structure is explicitly private and subject to change. This provides the missing API needed to determine whether a field is present.

### DIFF
--- a/test/com/walmartlabs/lacinia/selections_tests.clj
+++ b/test/com/walmartlabs/lacinia/selections_tests.clj
@@ -1,0 +1,83 @@
+(ns com.walmartlabs.lacinia.selections-tests
+  (:require
+    [clojure.test :refer [deftest is]]
+    [com.walmartlabs.lacinia.parser :as parser]
+    [com.walmartlabs.lacinia.executor :as executor]
+    [com.walmartlabs.test-utils :refer [default-schema]]
+    [com.walmartlabs.lacinia.constants :as constants]))
+
+(defn ^:private app-context
+  [query]
+  (let [parsed-query (parser/parse-query default-schema query)]
+    {constants/parsed-query-key parsed-query
+     constants/selection-key (-> parsed-query :selections first)}))
+
+
+(defn ^:private root-selections
+  [query]
+  (-> query
+      app-context
+      executor/selections-seq))
+
+(deftest simple-cases
+  (is (= [:character/name]
+         (root-selections "{ hero { name }}")))
+
+  (is (= [:human/name :human/homePlanet]
+         (root-selections "{ human { name homePlanet }}")))
+
+  (is (= [:human/name :human/friends :human/enemies
+          :character/name                                   ; friends
+          :character/appears_in
+          :character/name]                                  ; enemies
+         (root-selections "{ human { name friends { name appears_in } enemies { name }}}"))))
+
+(deftest mutations
+  (is (= [:human/name :human/homePlanet]
+         (root-selections "mutation { changeHeroHomePlanet (id: \"123\",
+            newHomePlanet: \"Venus\") {
+            name homePlanet
+          }
+         }"))))
+
+(deftest inline-fragments
+  (is (= [:character/name
+          :human/homePlanet
+          :droid/primary_function]
+         (root-selections
+           "{
+              hero {
+                name
+
+                ... on human { homePlanet }
+
+                ... on droid { primary_function }
+              }
+            }"))))
+
+(deftest named-fragments
+  (is (= [:character/name
+          :human/homePlanet
+          :droid/primary_function]
+         (root-selections
+           "query {
+              hero {
+                name
+
+                ... ifHuman
+                ... ifDroid
+              }
+            }
+
+            fragment ifHuman on human { homePlanet }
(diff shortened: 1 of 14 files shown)


test/com/walmartlabs/lacinia/selections_tests.clj
```diff
@@ -0,0 +1,83 @@
...
+  (is (= [:character/name]
+         (root-selections "{ hero { name }}")))
+
+  (is (= [:human/name :human/homePlanet]
+         (root-selections "{ human { name homePlanet }}")))
+
+  (is (= [:human/name :human/friends :human/enemies
+          :character/name                                   ; friends
+          :character/appears_in
+          :character/name]                                  ; enemies
```
sashton: Can the order of the returned selections be used for anything, or should the selections be returned as a set? It seems like it would be hard to match an ordered list of selections against the nested query to do anything useful with the ordering.

sashton: Here's an example which demonstrates that order can't be depended on to infer context:
```
(root-selections
 "{ human { ... on character {name} friends {appears_in}}}")
=> (:human/friends :character/name :character/appears_in)

(root-selections
 "{ human { friends {name appears_in}}}")
=> (:human/friends :character/name :character/appears_in)
```

hlship: If there was a lazy set, I would return that.  It's returned as an explicitly lazy seq, with a width (not depth) first spanning.  This means that checking for immediate children (e.g., :character/name from :QueryRoot/hero) is relatively cheap; with a set, then entire tree would have to be walked.

Not sure what you mean about ordering however; order shouldn't matter, and in many cases, fields will be resolved in parallel; the query order is just the order in which selections are assembled.

hlship: I think what you are looking for is a kind of reduction of the selection tree to something similar to the diagram above, from which more information can be extracted, at some cost.  I think that's a parallel strategy that can be put in place if it's needed; I believe that what we have so far is sufficient to address the needs of our query-service, and what I've gathered from chatting with users on the clojure/GraphQL slack channel.

hlship: Steve - a structural representation might be `[[:character/name] [:character/friends [:character/appears_in]]`.

sashton:
@hlship I'm trying to see how I could use this to solve the N+1 query problem. Compare the following two queries:
```
(root-selections
 "{ hero { name friends {name} enemies {name}}}")
=> (:character/name :character/friends :character/enemies :character/name :character/name)

(root-selections
 "{ hero { name friends {name enemies {name}}}}")
=> (:character/name :character/friends :character/name :character/enemies :character/name)
```
If I look at the selections, I can see that I'm dealing with a human, some friends, and some enemies. But I can't tell which entities need joining. Is it `human->enemies` or `human->friends->enemies`?

As another example:
```
{ hero { name friends {name enemies {name}}}}
{ hero { name enemies {name friends {name}}}}
```
and using this resolver (mentioned in the examples for this PR):
```
(defn resolve-hero
  [context args _]
  (if (executor/selects-field? context :character/friends)
    (fetch-hero-with-friends args)
    (fetch-hero args)))
```
`(executor/selects-field? context :character/friends)` will return true in both cases, but in the second case, we actually want to join hero with enemies, not friends.

hlship:
Cool that I was able to handle a trivial merge conflict here in the UI.

hlship:
Steve - absolutely; this is a first pass at solving problems we know exist.  I'm trying to get out the business of making grand predictions about how our software will be used. The structure representation I discussed earlier *might* address attempts to deal with N+1 fetches, but is certainly more awkward to use than `selects-field?` for common cases.

hlship:
Ok, I've added `selections-tree` and I think it addresses all the use cases we've come up with.

bcarrell:
I think I'm OK with this being merged, but I think I'd like to see this being filed under "experimental" (kind of like the timing feature recently added) and subject to feedback.

hlship:
Ready to merge?

PR #73, opened 2017-05-25
`merge` -> `deep-merge`

### LINKED ISSUE #52: Fields are not merged
This query:
```graphql
query { 
  viewer { 
    login
  }
  viewer {
    databaseId
  }
}
```

Should return something like:
```edn
{:data {:viewer {:login "someone" :id "qwertyuiop"}}}
```

Instead, it returns only the `databaseId`. If we query for the login last, it returns just the login.

bcarrell:
Minor change to ensure fields merge when they should.  Refer to the Issue for more details.

Closes #52

### DIFF
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -267,7 +267,7 @@
     (apply-selection execution-context selection)))
 
 (defn ^:private combine-map-results
-  "Left associative resolution of results, combined using merge."
+  "Left associative resolution of results, combined using deep-merge."
   [left-result right-result]
   (combine-results deep-merge left-result right-result))
 
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -2,7 +2,7 @@
   "Mechanisms for executing parsed queries against compiled schemas."
   (:require
     [com.walmartlabs.lacinia.internal-utils
-     :refer [cond-let to-message map-vals remove-vals q combine-results]]
+     :refer [cond-let to-message map-vals remove-vals q combine-results deep-merge]]
     [flatland.ordered.map :refer [ordered-map]]
     [com.walmartlabs.lacinia.schema :as schema]
     [com.walmartlabs.lacinia.resolve :as resolve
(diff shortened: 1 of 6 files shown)


test/com/walmartlabs/lacinia_test.clj
```diff
@@ -236,6 +254,16 @@
              }
            }"]
     (is (= {:data {:luke {:name "Luke Skywalker"}}}
+           (execute default-schema q nil nil))))
+  (let [q "query FetchLukeAliased {
+             luke: human(id: \"1000\") {
```
hlship: If I read the spec correctly, this is valid as long as all the arguments to `human` are the same. Do we check that? I think we need a validation and test to ensure that operation uniqueness is maintained.

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -267,9 +267,9 @@
     (apply-selection execution-context selection)))
 
 (defn ^:private combine-map-results
-  "Left associative resolution of results, combined using merge."
+  "Left associative resolution of results, combined using deep-merge."
```
hlship: There's some cases I'd be worried about that may not be handled by deep-merge, such as:

```
{ human { friends { name } }
  human { friends { id } } }
```

Which should produce `{:data {:human {:friends [{:name ... :id ...} ...]}}}`.  I'd like to see a test for that.

hlship:
I'm wondering if you explored merging together the field selections, rather than deep merging the data?  The former is complicated BUT would be more efficient, I suspect.  I'm going to pull this branch and run the benchmarks.

src/com/walmartlabs/lacinia/executor.clj
bcarrell: Yeah this doesn't work.  Fixing

bcarrell:
> I'm wondering if you explored merging together the field selections, rather than deep merging the data? The former is complicated BUT would be more efficient, I suspect. I'm going to pull this branch and run the benchmarks.

Looked briefly at doing it like this; it seemed like a complex way of going about it vs. merging the data.  Unsure the perf differences will be worth the effort and complexity.

test/com/walmartlabs/lacinia_test.clj
bcarrell: The case where the arguments are different isn't handled, at least as part of this PR.  This case should be a validation error (pre-execution).

hlship:
Ran the benchmarks, no appreciable difference in parse or execution time.

bcarrell:
Gonna close this for a bit while working on some stuff.

gusbicalho:
This solution will run the resolver multiple times, right?

bcarrell:
I'm going to explore doing this as part of parse instead of after execution has taken place to avoid the resolver issue.

PR #79, opened 2017-06-08
Merge selections for same fields (with same alias and arguments) together, as per spec

### LINKED ISSUE #52: Fields are not merged

hlship:
This is my approach to addressing this issue inside the parser; this does not require any execution changes.
 
 I'm going to check the performance impact, if any.

Fixes #52


hlship:
I can't think of many more possible tests but if you come up with any I'm glad to add them and ensure it all still works.

hlship:
I thought about this on the way home and there's very minor problems with this.  I have ideas for more tests, and for fixes.

hlship:
Happier with this. Collecting performance numbers right now.

test/com/walmartlabs/lacinia/merge_selections_test.clj
```diff
@@ -0,0 +1,137 @@
...
+}"))))
+
+(deftest understands-aliases
+  ;; Default for human is 1001 - Darth Vader
+  (is (= {:data {:darth {:name "Darth Vader"
+                         :homePlanet "Tatooine"}}}
+         (q "
+{
+  darth: human { name }
+  darth: human { homePlanet }
```
bcarrell: What if these two aliases are different (or one of them don't have an alias).  Are there two resolver calls?

hlship: Currently, two resolver calls.  I consider this to be an edge case, and it would be quite disruptive to the executor code to find a way to coalesce this into a single field resolver call.

bcarrell: @gusbicalho How do you feel about this case?

hlship: Do you know if the RI can do that optimization?

bcarrell: Nope, I'm unaware if it can.

hlship:
The CI failure isn't test failing, it's something in the CI server having trouble checking the code out.

test/com/walmartlabs/lacinia/merge_selections_test.clj
gusbicalho: To me it wouldn't be a problem to make two resolver calls in that case (same field with different aliases), seems to be a really edge case and shouldn't happen in my use case (relay)

hlship:
Yes, there's still some edge cases not covered so well. For example, there's something about fragments defining aliases that are conflicting with the main selection (or perhaps, each other).  There are definitely some parts of the spec that are defined in terms of whats convenient for the RI to implement.

PR #83, opened 2017-06-16
Fix NullPointerException in Selections API

bcarrell:
Queries that include introspective fields like `__typeName` are currently throwing NullPointerExceptions when using the selections API because they don't have `type-name` or `field-name`.

It's difficult to establish what the type-name or field-name should be for these meta style fields (moreso the type-name), so let's just guard against nils in the selections API for now.

### DIFF
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -446,10 +446,11 @@
 
 (defn ^:private to-field-name
   [node]
-  (let [field-def (:field-definition node)]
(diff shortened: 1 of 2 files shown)


hlship:
Would another option be to simply edit out __typename (i.e., field selection w/o a field)?  It doesn't actually provide value in the selections api, which should really be just about the application's fields, not the libraries.

hlship:
Please add a test.

hlship:
Doe this scenario affect `selections-tree`?  Again, a test would be nice.

PR #84, opened 2017-06-16
Add support for GraphQL subscriptions

### LINKED ISSUE #78: Subscription support
We're getting closer to adding subscription support, which has, or is about to, hit the official spec.

Here's my initial thoughts on implementing subscriptions:

For subscription operations, there's *two* callbacks:  the traditional field resolver, but also a subscriber function, which I'm tentatively calling `stream`.

The stream function is passed the application context, the operation's arguments, and an event function.

The stream function has two responsibilities:
- Provide resolved values to the event function
- Return a cleanup callback

The subscriber is responsible for invoking the event function when there's new data to send to the subscribed client.

In other words, the stream function is responsible
for repeatedly invoking the event function with a succession of values, as they become available.  Presumably, there will be a started thread, or core.async CSP, responsible for this.

The value passed to the event function becomes the resolved value passed to the field resolver (rather than the nil value normally passed to a top level operation).

After invoking the event function, the provided value will be passed to the fields resolver, much as with a traditional query operation.

There is no guarantee that the resolver will be invoked in the same thread. More likely, there will be some amount of asynchronous processing going on.

This continues until the client disconnects, or the event function is invoked with nil.

The stream function returns a cleanup function; this function is invoked after the subscription is canceled (weather it was closed by passing nil to the event function, by the client closing the connection, or by the loss of the connection to the client).

hlship:
This adds code, tests, and documentation.  After this comes an implementation in lacinia-pedestal.

Fixes #78


hlship:
Ok, I'll fix the conflicts.

hlship:
Ok, pretty easy merge.

hlship:
Hiccups on Circle CI.  All tests pass, you may need to pull them and execute local if you want.

src/com/walmartlabs/lacinia.clj
```diff
@@ -38,9 +38,7 @@
     (resolve/resolve-as {:errors validation-errors})
 
     :else
-    (executor/execute-query (assoc context
-                                   constants/schema-key schema
```
hlship: The parsed query already contains the schema, no need to add it again.

hlship: This is predicting changes in lacinia-pedestal related to executing subscriptions as a query when values arrive on the source stream.

docs/subscriptions/overview.rst
```diff
@@ -0,0 +1,30 @@
...
+- Lacinia invokes a streamer function once, to initialize the subscription stream.
+
+- The streamer is provided with a source stream callback function; as new values are available
+  they are passed to this callback.
+
+  Typically, the streamer will create a thread, core.async process, or other long-lived
+  construct to feed values to the source stream.
+
+- Whenever the source stream callback is passed a value,
+  Lacinia will execute the the subscription as a query, which will generate a
```
annapawlicka: repeated `the` that should be removed

docs/subscriptions/resolver.rst
```diff
@@ -0,0 +1,16 @@
...
+Unlike a query or mutation, the :doc:`field resolver <../resolve/index>`
+for a subscription always starts with a specific value, provided by the streamer, via
+the source stream callback.
+
+Because of this, the resolver is optional: if not provided, a default resolver is used, one that simply returns
+the source stream value.
+
+However, it still makes sense to implement a resolver in some cases.
+
+Both the resolver and the streamer receive the same map of arguments: it is reasonble that some
```
annapawlicka: typo, should be `reasonable`

docs/subscriptions/streamer.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. literalinclude:: ../_examples/subs-streamer.edn
+  :language: clojure
+
+A lot of this example is hypothetical; it presumes ``create-log-subscription`` will return an value
+that can be used with ``on-publish`` and ``stop-log-subscription``.
+A real implementation might use Clojure core.async, subscribe to a JMS queue, or an almost
+unbounded number of other options.
+
+Regardless, the streamer provides the stream of source values, but making successive calls to
```
annapawlicka: s/but/by

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -253,6 +254,10 @@
 
 ;; This can be expanded at some point
 (s/def :type/type some?)
+(s/def :type/arg (s/keys :req-un [:type/type]
+                         :opt-un [:type/description]))
+(s/def :type/args (s/map-of keyword? :type/arg))
+;; TODO: No longer accurate, :resolve must always be a function is present.
```
annapawlicka: Is this TODO for another PR or should be done as part of this one?

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -7,7 +7,8 @@
     [com.walmartlabs.lacinia.schema :as schema]
     [com.walmartlabs.lacinia.resolve :as resolve
      :refer [resolve-as]]
-    [com.walmartlabs.lacinia.constants :as constants])
+    [com.walmartlabs.lacinia.constants :as constants]
+    [clojure.spec.alpha :as s])
```
annapawlicka: This doesn't seem to be used in this ns.

test/com/walmartlabs/lacinia/arguments_test.clj
```diff
@@ -1,4 +1,4 @@
-(ns com.walmartlabs.lacinia.arguments-test
+ (ns com.walmartlabs.lacinia.arguments-test
```
annapawlicka: Whitespace

test/com/walmartlabs/lacinia/subscription_tests.clj
```diff
@@ -0,0 +1,158 @@
...
+  (is (nil? (latest-response))))
+
+(deftest ensure-variables-and-arguments-are-resolved
+  (execute "
+subscription ($severity : String) {
+
+  logs (severity: $severity) {
+     severity message
+  }
+}" {:severity "critical"})
```
annapawlicka: Can we re-format this string query?

test/com/walmartlabs/lacinia/subscription_tests.clj
```diff
@@ -0,0 +1,158 @@
...
+  (log-event nil)
+
+  (is (nil? (latest-response))))
+
+(deftest one-subscription-per-request
+  (when-let [e (is (thrown? Exception (parser/parse-query compiled-schema "
+subscription {
+  sev: logs { severity }
+  msg: logs { message }
+}")))]
```
annapawlicka: Re-format the string?

src/com/walmartlabs/lacinia/schema.clj
hlship: It's not essential, so it can be deferred.

src/com/walmartlabs/lacinia/executor.clj
hlship: It was for a while ... then I tore it out.

test/com/walmartlabs/lacinia/subscription_tests.clj
hlship: Why?

annapawlicka: Because when you just copy and paste it, it looks the way it does now: out of place and harder to read.

src/com/walmartlabs/lacinia/introspection.clj
```diff
@@ -75,6 +75,8 @@
                          (update :fields #(remove-keys is-internal-type-name? %)))
         mutations-root (get schema constants/mutation-root)
         omit-mutations (-> mutations-root :fields empty?)
+        subs-root (get schema constants/subscription-root)
+        omit-subs (-> subs-root :fields empty?)
         type-names' (if omit-mutations
                       (-> type-names set (disj constants/mutation-root))
```
hlship: Do we know if it is required to omit mutations and subscriptions if they have no fields?  This should probably be fixed to do the same thing for subscriptions that is done for mutations.

hlship: I changed it to be consistent.

annapawlicka:
Conflicts.

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -289,13 +294,42 @@
...
+
+(s/def :type/subscription (s/keys :opt-un [:type/description
+                                           :type/resolve
+                                           :type/args]
+                                  :req-un [:type/type
+                                           :type/stream]))
+
+(s/def :type/subscriptions (s/map-of keyword? :type/subscription))
+
 (s/def ::schema-object
```
hlship: So I'm thinking that we should move all the specs into the com.walmartlabs.lacinia.schema space, like ::schema-object.

bcarrell:
Seems simple enough -- ship it!

PR #89, opened 2017-06-30
Add ability for field resolvers to modify the application context passed to nested field resolvers

hlship:
This change hits a few things:

- The selections pipeline has been modified using the parameters design pattern (a map of values, rather than the previous 3 parameters)
- RenderCommand protocol has been added: it defines a wrapper type that can modify the execution context
- ResolverResult is now about a single value, and there's a `with-error` wrapper function that wraps a resolved value with a RenderCommand
- The callback for `on-resolve!` now gets just a single value
- Logic in `executor` now applies any RenderCommands on the resolved value *before* passing it down the selector pipeline

... all that is to support:

- `with-context` wraps a resolved value such that nested field resolvers get a modified context

Also, I found that I can get reasonably accurate performance results using `quick-benchmark` rather than `benchmark`, and save a lot of time.


hlship:
Closing until I see about reverting just the selector pipeline changes.

hlship:
Ok, I made a few improvements and am now happy with it.  There's been a tiny uptick in execution time, but only in the very large queries (such as introspection).

hlship:
Circle Ci is doing that thing again; tests do pass!

docs/resolve/context.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Here, the search term is provided to the ``products`` field, but is again needed by the ``highlighted_name``
+field, to highlight the parts of the name that match the search term.
+
+The resolver for the ``product`` field can communicate this "down tree" to the resolver for the ``highlighted_name`` field:
+
+.. literalinclude:: /_examples/mutable-context.edn
+   :language: clojure
+
+This new key, ``::search-term`` is only present in the context below ``products`` field.
```
candid82: Could this be solved by adding `::search-term` in `product` resolver's result value without modifying the context?

hlship: I put the `category` field in between them to justify the use of the context change.

candid82: Ah, I missed that part. Makes sense!

docs/resolve/context.rst
```diff
@@ -0,0 +1,34 @@
+Application Context
+===================
+
+The application context passed to your field resolvers is normally set by the initial call to
+`com.walmartlabs.lacina/execute-query`.
+Lacinia uses the context for its own book-keeping (they keys it places into the map are namespaced to
```
sashton: typo: `they`

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -44,14 +43,13 @@
...
-  each error, including location and arguments."
-  [field-selection error-maps]
-  (when (seq error-maps)
-    (let [enhanced-data (ex-info-map field-selection nil)]
-      (map
-        #(merge % enhanced-data)
-        error-maps))))
+  "From an error map, or a collection of error maps, add additional data to
+  each error, including location and arguments.  Returns a seq of error maps."
+  [field-selection error]
```
sashton: should `error` be renamed `errors` to match the function name?

docs/resolve/context.rst
hlship: ok

src/com/walmartlabs/lacinia/executor.clj
hlship: it's that darned `error-map-or-seq-of-error-maps` flexibility.

src/com/walmartlabs/lacinia/resolve.clj
```diff
@@ -25,63 +61,79 @@
...
 
 (defn resolve-promise
   "Returns a ResolverResultPromise.
 
    A value must be resolved and ultimately provided via [[deliver!]]."
   []
-  (let [realized-result (promise)
-        callback-promise (promise)]
+  (let [*result (atom ::unset)
+        *callback (atom nil)]
```
sashton: Why the change from `promise`s to `atom`s?

hlship: I peeked inside the implementation of promise and found ... an atom.  Which surprised me. If I was really concerned about race conditions, probably use some java.util.concurrent stuff ... but the use case for all this is one thread perhaps creates the ResolverResultPromise and hands it off to another thread that delivers the value.  It would probably be nice to make this 100% bullet proof, but I don't see it as urgent.

test/com/walmartlabs/lacinia/resolver_result_test.clj
```diff
@@ -12,32 +12,29 @@
 
 (deftest callback-is-invoked-with-value-and-errors
```
sashton: I don't see the `-and-errors` part of this test anymore.

src/com/walmartlabs/lacinia/resolve.clj
sashton: In this case, the semantics of `promise/deliver/realized?` feel to me to match the use case in the code, and I find the promise version easier to read. Is there a performance improvement in the `atom` version?

hlship: No, and I think you are right about the readability. I've moved it back towards using promises.

hlship:
I believe all review issues have been addressed.

PR #90, opened 2017-07-04
Ensure that all identifiers in the schema are valid GraphQL identifiers

### LINKED ISSUE #88: Schema compiler should enforce that field and entity names are properly formed
For example:
`(execute my-schema "queryname { some-field-with-dashes }" nil nil)`
return following error:
`extraneous input '-' expecting {'(', '{', '}', '...', '@', Name}`

While my schema allow me to have fields with dashes - i.e. this is only query time issue. I managed to track this down to file: `com\walmartlabs\lacinia\Graphql.g4` line 152
```
Name
    : [_A-Za-z][_0-9A-Za-z]*
    ;
```

I think this regexp should look like this rather:
```
Name
    : [_A-Za-z-][_0-9A-Za-z-]*
    ;
```
i.e. allow dash as part of field name.

hlship:
Validate that all ids are valid GraphQL identifiers

- objects, unions, interfaces, etc.
- field names
- argument names
- enum values

Fixes #88


bcarrell:
Conflicts

hlship:
Conflicts resolved.

PR #91, opened 2017-07-04
Add CircleCI 2.0 config

ProjectFrank:
This config enables the use of CircleCI 2.0, which grants an increase in reliability and a decrease in build time from 2 minutes+ to about 12 seconds.


bcarrell:
Sweet, thanks!

PR #92, opened 2017-07-05
Cleanups for specs in com.walmartlabs.lacinia.schema

### LINKED ISSUE #81: spec for GraphQL schema does not define :queries and :mutations
The spec definitions in com.walmartlabs.lacinia.schema does not provide any detail for the :queries and :mutations keys.  These are `(s/map-of keyword? :type/operation)`, where an operation is just about the same as a field, but :resolve should be required (not optional).

hlship:
This PR addresses a few issues to the use of spec

- Moves the specs into the com.walmartlabs.lacinia.schema namespace
- Explicitly requires the use of simple keywords for keys in the input schema
- Defines :queries and :mutations
- Removes an outdated definition of the :resolve key on a field
- Properly defines the ::type spec (which is recursive!)

Fixes #81

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -258,8 +258,8 @@
 ;; ## Validations
 
 (s/def ::type (s/or :base-type schema-reference?
-                    :complex-type (s/cat :modifier #{'list 'non-null}
-                                         :type ::type)))
+                    :wrapping-type (s/cat :modifier #{'list 'non-null}
+                                          :type ::type)))
 (s/def ::arg (s/keys :req-un [::type]
                      :opt-un [::description]))
 (s/def ::args (s/map-of simple-keyword? ::arg))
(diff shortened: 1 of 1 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -250,89 +257,97 @@
...
-(s/def :type/fields (s/map-of keyword? :type/field))
-(s/def :type/implements (s/coll-of keyword?))
-(s/def :type/description string?)
-(s/def :type/object (s/keys :req-un [:type/fields]
-                            :opt-un [:type/implements
-                                     :type/description]))
-(s/def :type/interface (s/keys :opt-un [:type/description
-                                        :type/fields]))
+(s/def ::type (s/or :base-type schema-reference?
+                    :complex-type (s/cat :modifier #{'list 'non-null}
```
bcarrell: I think this definition should refer to wrapping type by renaming `complex-type` to `wrapping-type` to be congruent with the specification:

> The type system might want to define that it returns a list of other types; the List type is provided for this reason, and wraps another type. Similarly, the Non-Null type wraps another type, and denotes that the result will never be null. These two types are referred to as “wrapping types”; non‐wrapping types are referred to as “base types”. A wrapping type has an underlying “base type”, found by continually unwrapping the type until a base type is found.

hlship: Ok, sounds great.

PR #93, opened 2017-07-05
Ensure that every query and mutation has a resolver function

hlship:
This is something that is hard to catch via spec, and we can't be sure the application is using spec (they should!).

This adds checks that :resolve is present AND a function.

Fixes #90  ... but this builds on PR #90

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,9 +17,8 @@ changed to only accept a single value.
 Previously, the callback received a resolved value and an nilable error map.
 This method is not intended for use in application code.
 
-The `com.walmartlabs.lacinia.schema/compile` function is now *always* instrumented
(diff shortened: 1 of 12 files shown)


hlship:
I'll do a master merge after the other PRs are merged.

hlship:
Ok, done.  However, I have an idea where we could shift even more of this kind of validation over to spec.

Hm. Let me check that build failure.

hlship:
The build works fine locally, both `lein test` and `lein test2junit`, not sure why Circle is having a problem. Will ssh in later to see.

hlship:
Ok, ready for review. This grew a bit in scope, but it ultimately means less code in Lacinia, by leaning on spec.

PR #96, opened 2017-07-18
Simplify validation and hide the namespaces

hlship:
Validation is not, at least currently, extensible, or intended for application code to invoke. This PR changes those namespaces to be invisible in the documentation.

The signatures of the validation functions are a bit out of date; the parsed and prepared query has everything needed, including the full schema if that is required.  The functions now simply receive the prepared schema.

### DIFF
--- a/src/com/walmartlabs/lacinia.clj
+++ b/src/com/walmartlabs/lacinia.clj
@@ -23,8 +23,7 @@
          (or (nil? context)
              (map? context))]}
   (cond-let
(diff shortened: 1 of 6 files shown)


candid82:
I am not a fan of `' quotes, but as I understand it's already used in many places across lacinia codebase, and making things consistent is good.  
Tests failed, but otherwise lgtm.

bcarrell:
Tests pass locally

bcarrell:
retest

Issue #30, opened 2017-04-04
List types should allow a single value on input


hlship:
Noticed this in the [spec](http://facebook.github.io/graphql/#sec-Lists):

```
When expected as an input, list values are accepted only when each item in the list can be accepted by the list’s item type.

If the value passed as an input to a list type is not a list and not the null value, it should be coerced as though the input was a list of size one, where the value passed is the only item in the list. This is to allow inputs that accept a “var args” to declare their input type as a list; if only one argument is passed (a common case), the client can just pass that value rather than constructing the list.

that when a null value is provided via a runtime variable value for a list type, the value is interpreted as no list being provided, and not a list of size one with the value null.
```

I believe the code is currently more strict on this item, enforcing that lists passed as input values (e.g., field argument, field of input object, or variable) are in fact lists.

annapawlicka:
I'm going to look into this one.

annapawlicka:
Resolved by https://github.com/walmartlabs/lacinia/pull/97

PR #97, opened 2017-07-19
[20170717]  Allow single values on list type input

annapawlicka:
Fixes issue #30 

Field arguments, fields of input objects and variables of type list will now allow input of a single value and will coerce it to a list of size one.


annapawlicka:
I need to remove Pivotal Tracker hook from my Emacs config..

hlship:
I'm back and will review this shortly.

PR #101, opened 2017-07-25
Allow ResolverResultPromise callbacks to execute in a thread pool

hlship:
My concern is that when a mix of normal and async resolvers are operating inside core.async, a `go` block that resolves a promise could trigger a cascade of work that may include blocking I/O and lead to thread pool starvation.

### DIFF
--- a/src/com/walmartlabs/lacinia/resolve.clj
+++ b/src/com/walmartlabs/lacinia/resolve.clj
@@ -1,5 +1,12 @@
 (ns com.walmartlabs.lacinia.resolve
-  "Complex results for field resolver functions.")
+  "Complex results for field resolver functions."
+  (:import (java.util.concurrent Executor)))
+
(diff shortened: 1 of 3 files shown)


src/com/walmartlabs/lacinia/resolve.clj
```diff
@@ -119,7 +130,10 @@
         (deliver *result resolved-value)
 
         (when (realized? *callback)
-          (@*callback resolved-value))
+          (let [callback @*callback]
+            (if-some [^Executor executor *callback-executor*]
```
sashton: Is there a reason for `if-some` vs the more common `if-let`?

hlship: Just wanted to be explicit about it being a nil check though a callback should be a function, not a boolean.

Issue #20, opened 2017-03-27
Variable nil vs. missing variable for field argument


hlship:
From the GraphQL spec, concerning [null values](http://facebook.github.io/graphql/#sec-Null-Value)

```
The same two methods of representing the lack of a value are possible via variables by either providing the a variable value as null and not providing a variable value at all.
```

I'm pretty sure that the implementation will pass a nil for the argument value if the argument is bound to a variable that is simply not provided in the query.  If so, this must be changed.

annapawlicka:
Looking into it.

annapawlicka:
Resolved in https://github.com/walmartlabs/lacinia/pull/97

annapawlicka:
Fixed with #102

PR #102, opened 2017-07-25
Remove arguments bound to a variable that has no value provided in a query

annapawlicka:
This PR fixes issue #20 
+ Remove arguments bound to a variable that has no value provided in a query
  + null values, e.g. `foo(bar: null)` are values and are not removed
+ Update test schema
  + new tests to test missing/null variables
  + changed resolver for `hero` query. Previously if nullable argument `episode` was not provided, the resolver would still provide a value even though `default-value` was not specified. This behavior would mask an error about non-nullable character (`hero`) being null. Quite a few tests had to be updated to add `episode` argument (their logic has not changed) but I think this new behavior makes more sense.

### DIFF
--- a/dev-resources/com/walmartlabs/test_schema.clj
+++ b/dev-resources/com/walmartlabs/test_schema.clj
@@ -128,6 +128,15 @@
                    "2001"                                   ; r2d2
                    )))
 
+(defn get-villain
+  "Retrieves the biggest villain of the episode."
+  [episode]
+  (get-character (condp = episode
+                   :NEWHOPE "1004" ;; Tarkin
+                   :EMPIRE "1001" ;; Vader
+                   :JEDI "1001"
+                   nil)))
+
 ;; —————————————————————————————————————————————————————————————————————————————
 ;; ## Schema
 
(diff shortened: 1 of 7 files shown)


dev-resources/com/walmartlabs/test_schema.clj
```diff
@@ -233,6 +242,43 @@
                             :resolve (fn [_ _ _]
                                        "<Single Value>")}}}
 
+    :villain
```
hlship: I know this is out of the blue but ... I really think that putting more code into test_schema.clj is a problem; I've been adding lots of more narrowly focused tests and namespaces and I feel that is really the right way to go.

hlship: For example, each change here forces a change to the introspection tests.

annapawlicka: Working on moving new test data and tests to a new namespace.

annapawlicka:
@hlship I removed new object/query from test_schema.clj and updated separate variable tests.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -629,7 +629,13 @@
         ;; This is kind of a juxt buried in a map. Each value is a function that accepts
         ;; the variables and returns the actual value to use.
         (fn [variables]
-          (map-vals #(% variables) dynamic-args))))))
+          (->> (map-vals #(% variables) dynamic-args)
+               (filter (fn [[variable value]]
+                         (let [[_ user-defined-var-name] (get arguments variable)]
+                           ;; keep arguments that have a matching variable provided
+                           (or (contains? variables user-defined-var-name)
+                               ;; or have a default-value specified
```
hlship: I think there's a tiny bit of ambiguity here between a default value of nil or the lack of a default value at all.  This code treats a default value of nil as a lack of default value.  However, I'm willing to let that slide until/unless that becomes an issue.

annapawlicka:
Right, I can remove the changes I made to `hero` query too, considering we want to move away from using test-schema.

src/com/walmartlabs/lacinia/parser.clj
annapawlicka: I added a test to catch this scenario. Will try to fix it in this PR.

annapawlicka:
I'm closing this PR while I'm making more changes.

annapawlicka:
This PR is now ready for a review, @hlship @bcarrell

src/com/walmartlabs/lacinia/internal_utils.clj
```diff
@@ -57,6 +57,19 @@
         persistent!
         (with-meta (meta m)))))
 
+(defn filter-kvs
+  "Builds a new map passing each key and value to predicate."
```
hlship: The predicate only seems to involve the `key`, not the `value`.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -629,7 +633,12 @@
         ;; This is kind of a juxt buried in a map. Each value is a function that accepts
         ;; the variables and returns the actual value to use.
         (fn [variables]
-          (map-vals #(% variables) dynamic-args))))))
+          (->> (map-vals #(% variables) dynamic-args)
+               ;; keep arguments that have a matching variable provided.
+               ;; :value in value-map might be NULL but it's still a
+               ;; provided value (e.g. may be used to indicate deletion)
+               (filter-kvs (fn [variable value-map] (some? value-map)))
+               (map-vals (fn [v] (:value v)))))))))
```
hlship: `(map-vals :value)` would work here

src/com/walmartlabs/lacinia/internal_utils.clj
annapawlicka: Yep. It didn't before I refactored some of the stuff. Will fix it.

hlship: Actually, filter-keys is what this is doing!

annapawlicka:
Ok, I think it's much simpler now @hlship

Issue #17, opened 2017-03-27
Fields with arguments are not checked against fields in interface


hlship:
We have good checks in place that a field in an object uses the same, or compatible, types as the field in an interface from which the object inherits.

However, there is no check that the interface field and the object field have the same arguments.

The interface and field should define the same set of arguments, with identical types (because argument types are limited, there's no question of unions, interfaces, or objects: just scalar types, enums, and input objects.

annapawlicka:
Looking into it.

annapawlicka:
Fixed in #106

PR #106, opened 2017-08-02
Check that arguments on the field are compatible with arguments on the interface field

annapawlicka:
This PR fixes issue #17 

>The object field must include an argument of the same name for every argument defined in the interface field.
The object field may include additional arguments not defined in the interface field, but any additional argument must not be required.

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -1015,7 +1015,7 @@
       (doseq [additional-arg-name additional-args
               :let [arg-kind (get-in object-field-args [additional-arg-name :type :kind])]]
         (when (= arg-kind :non-null)
-          (throw (ex-info "Additional arguments on an object field cannot be required."
+          (throw (ex-info "Additional arguments on an object field that are not defined in extended interface cannot be required."
                           {:object type-name
                            :interface-name interface-name
                            :field-name field-name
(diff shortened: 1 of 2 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -971,24 +971,43 @@
...
+                           :field-name field-name
+                           :interface-name interface-name
+                           :argument-name arg-name})))
+
+        (when-not (is-assignable? schema interface-arg-def object-field-arg-def)
+          (throw (ex-info "Object field's argument is not compatible with extended interface's argument type."
+                          {:object type-name
+                           :interface-name interface-name
+                           :field-name field-name
+                           :argument-name arg-name}))))))
```
hlship: I'm not seeing the code that checks that any additional non-interface-defined arguments are optional.

annapawlicka:
@hlship Added in https://github.com/walmartlabs/lacinia/pull/106/commits/833e97ae86b5f56f9845c542b261d8af52f3e4d2

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -1007,7 +1008,18 @@
...
                            :field-name field-name
-                           :argument-name arg-name}))))))
+                           :argument-name arg-name})))))
+
+    (when-let [additional-args (seq (difference (into #{} (keys object-field-args))
+                                                (into #{} (keys interface-field-args))))]
+      (doseq [additional-arg-name additional-args
+              :let [arg-kind (get-in object-field-args [additional-arg-name :type :kind])]]
+        (when (= arg-kind :non-null)
+          (throw (ex-info "Additional arguments on an object field cannot be required."
```
hlship: I think the error message needs to mention that an interface field and argument is involved.

PR #107, opened 2017-08-09
Make a number of small performance optimizations

hlship:
These optimizations are targeted at very large lists with sub-selections.


hlship:
[exec-time.pdf](https://github.com/walmartlabs/lacinia/files/1215946/exec-time.pdf)

The changes seem to be improving the larger benchmarks, but not in a significant way.

src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -1050,8 +1052,9 @@
   "The default for the :default-field-resolver option, this uses the field name as the key into
   the resolved value."
   [field-name]
-  ^ResolverResultImpl
```
hlship: This was actually a bug that had performance impact!

project.clj
```diff
@@ -1,18 +1,18 @@
-(defproject com.walmartlabs/lacinia "0.20.0"
+(defproject com.walmartlabs/lacinia "0.21.0"
   :description "A GraphQL server implementation in Clojure"
   :license {:name "Apache, Version 2.0"
             :url "http://www.apache.org/licenses/LICENSE-2.0"}
   :plugins [[lein-codox "0.10.3"]
             [test2junit "1.2.5"]]
   :dependencies [[org.clojure/clojure "1.8.0"]
                  [clojure-future-spec "1.9.0-alpha17"]
+                 [org.clojure/test.check "0.9.0"]
```
thorstadt: Does test.check need to be a run-time dependency?

hlship: I believe so because we *always* instrument c.w.lacinia.schema/compile.  This turned up as a problem when doing to upgrade to Lacinia 0.20.0 in eReceipts.

hlship: But looking again at the code, I'm not sure *why* test.check is needed just because of instrument.  I'm going to move this back and figure it out in a separate PR.

bcarrell:
conflicts after merging the others

Issue #108, opened 2017-08-17
String escaping doesn't work


mullr:
When passing a string as an argument to a query, internal escaping isn't evaluated. 

I wrote a test case here: https://github.com/mullr/lacinia/tree/string-escaping-repro

mullr:
More specifically: https://github.com/mullr/lacinia/blob/string-escaping-repro/test/com/walmartlabs/lacinia/parser_test.clj#L44

thorstadt:
Fixed in #111

baldmountain:
I don't think this is fixed. #111 is isolated to query parsing. The example above is a mutation. I'm currently working around this by url encoding the parameter field (a JSON string filled with double quotes) in the client and url decoding it in the server.

thorstadt:
@baldmountain Lacinia parses queries and mutations the same way, the PR includes a test for this case: https://github.com/walmartlabs/lacinia/pull/111/files#diff-3d2e5adacfff4c8484d5c3a5930edbbeR48

0.21.0, released yesterday, is the first to contain this fix. Are you using that version?

baldmountain:
I'm using lacinia-pedestal so that will need the update from 0.20.0 to o.21.0.

PR #111, opened 2017-08-24
String escaping

thorstadt:
This is a proposed fix for #108, adding ASCII and Unicode escape sequence parsing for String arguments and variables.

### DIFF
--- a/test/com/walmartlabs/lacinia/variables_test.clj
+++ b/test/com/walmartlabs/lacinia/variables_test.clj
@@ -130,6 +130,19 @@
     (schema/tag-with-type v type)
     v))
 
+(deftest variables-with-escape-sequences
+  (let [value (atom nil)
+        schema (schema/compile
+                {:objects {:value {:fields {:value {:type 'String}}}}
+                 :mutations {:change {:type :value
+                                      :args {:newValue {:type 'String}}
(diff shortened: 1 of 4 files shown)


test/com/walmartlabs/lacinia/variables_test.clj
```diff
@@ -130,6 +130,19 @@
...
+  (let [value (atom nil)
+        schema (schema/compile
+                {:objects {:value {:fields {:value {:type 'String}}}}
+                 :mutations {:change {:type :value
+                                      :args {:newValue {:type 'String}}
+                                      :resolve (fn [ctx args v]
+                                                 (reset! value (:newValue args))
+                                                 {:value @value})}}})
+        q (parser/parse-query schema "mutation ($ValueArg: String) { change(newValue: $ValueArg) { value }}")]
+    (is (= (execute-parsed-query q {:ValueArg "\\\"\\u2B50\\\""} nil)
```
hlship: I'm not happy with this.  For the query document, which we get as a string, we should be responsible for building the correct final value of embedded strings.  However, for arguments, that should be the responsibility of the JSON parser.

thorstadt:
@hlship Good point -- see e80c554. I've isolated String unescaping to query parsing.

PR #117, opened 2017-10-03
Deep merge fragments/fields

gusbicalho:
I've run into a few more problems with selections from several fragments which should be merged, but aren't. I've include a few tests that show the kind of cases we should be able to handle, and you can see they either pass or fail depending on the order of the fragments.

I've tried my hand at fixing the problem, and was able to merge selections for inline fragments on the same concrete-type, but even fragments on different types could run into this issue. And that still leaves out the problem of named fragments and merging between fragments and fields.

I won't be able to pursue this in the near future, so if the mantainers or anyone wants to take a shot at solving this, please do!

### DIFF
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -871,12 +871,16 @@
   "The selection key only applies to fields (not fragments) and
   consists of the field name or alias, and the arguments."
   [selection]
-  (let [{:keys [:selection-type :alias]} selection]
-    (if (= selection-type :field)
+  (let [{:keys [:selection-type :alias :concrete-types :fragment-name]} selection]
+    (case selection-type
+      :field
       alias
-      ;; TODO: This may be too simplified ... worried about loss of data when merging things together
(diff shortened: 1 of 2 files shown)


hlship:
Not completely ignoring this; it's a bit of a challenge to fix in a unified way without affecting performance, so I'm thinking about it in the background as I clear out the more low hanging fruit.

gusbicalho:
Ok @hlship, thanks anyway for the great work on this lib!

hlship:
Glad you like it!  I wish some of the other internal team members represented themselves better outside of Walmart. They seem to leave all the communication and evangelism to me, which is fine, but leaves the mistaken impression that this is a one-person project.

hlship:
I haven't forgotten about this, still trying to figure out the right approach.

hlship:
I think so. I"m curious how/if other implementations deal with this. Possibly we need to "flag" parts of the sub-tree as needing a merge phase, after selections.  Or possibly there's a way to split up the selections inside a fragment in a way that works.

PR #122, opened 2017-10-24
Fix bug (related to #114) that caused named operations to not work

hlship:
A trip up related to the code the ugly support for 'query' etc. as a operation/field/type name.


hlship:
I've been thinking in terms of an intermediate format between the Antlr parse tree, and the tree that is converted into executable state ... first, it could iron out some of this ugliness and second it could make it possible to support an alternate parser (e.g., something in JavaScript to allow Lacinia to run inside Node).

PR #123, opened 2017-10-24
API to wrap a resolver function with a new function that can modify or replace the resolved value

hlship:
This is something that its going to be very useful for folks that want to wrap very Clojure-oriented data structures (such as out from Datomic) and do the conversions to JSON camel case on the fly.  I'm sure there are other applications. The solution is a bit dense, because of the potential variations:

- The resolver function returns a simple value
- The resolver function returns a ResolverResult
- The returned value has been decorated via with-error or with-context
- The wrapped value is a simple value
- The wrapped value is a ResolverResult

0.22.1 is out, so this can be merged to master (0.23.0).


annapawlicka:
Needs merge from master.

PR #130, opened 2017-10-28
Change list type check to sequential? to allow syntax quoted list types

thosmos:
This change enables generating schema type definitions using syntax quoted lists like so:
```clojure
:queries
   (into {}
     (for [f [:one :two :three]]
       [f {:type `(~'list ~f)
           :resolve :generated-resolver}]))
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject com.walmartlabs/lacinia "0.22.1"
+(defproject com.walmartlabs/lacinia "0.22.1-SNAPSHOT"
   :description "A GraphQL server implementation in Clojure"
   :license {:name "Apache, Version 2.0"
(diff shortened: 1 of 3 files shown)


hlship:
Although we generally just load a fixed schema from EDN, we also make a big point about it being "just data", from any source, transformable before compilation ... so I can see this as being useful.

thosmos:
Oops, I didn't mean to include the version change in this PR.

hlship:
If you could push an update that doesn't touch the version number, that would be best.

Version numbers are tricky, because depending on what we decide to include in a release affects what the version number should be.  We're trying to avoid a more complicated branch structure (we prefer to keep it just `master`, and PR branches).

PR #135, opened 2017-11-27
Prevent an exception calling ex-info that can (rarely) occur because of nil ex-data

xfthhxk:
Clojure's `ExceptionInfo` constructor expects the data map passed to it to be non-null.  I encountered a situation where all the arguments to `merge` where `nil` which returns `nil`.  The actual Lacinia exception message is lost and you see  "Additional data must be non-nil."


hlship:
Looks fine, and trivial enough that I'm good without a test case.

Merge will be after the 0.23.0 release.

PR #138, opened 2017-12-01
Add API to allow ResolverResults to be chained together

hlship:
This adds function `chain*` and macro `chain` that make it possible to chain together the value from a ResolverResult forming new ResolverResults; it re-uses code and logic from a the previous `wrap-resolver-result` function.

Although they make the code cleaner, when I made use of them inside `executor`, there was a noticeable effect on performance ... likely due to the work necessary to unpack and repack ResolveCommands, along with extra checks that returned values are ResolverResults ... the executor code doesn't need those extra checks and they add up.

I still think these functions have use, just outside of the core of Lacinia.


hlship:
From the discussion; may preview these internally and add them, if needed, later.

Issue #140, opened 2017-12-03
Enum is not enforced in an object field


itaied246:
Maybe I didn't understand it correctly, or the `enum` is not working as expected.

I have this `enum` type:

```edn
{:platform
 {:description "Supported gaming platforms"
  :values      [:PLAYSTATION_4 :PLAYSTATION_3 :XBOX_ONE :XBOX_360 :PC]}}
```

And this `object`:

```edn
 :game
 {:fields
  {:name     {:type String}
   :platform {:type :platform}}}
```

As I understand `enum`s, only the values specified are allowed.
But if I return the following result:

```clj
:query/post-by-id (fn [context args value]
                       {:game {:name     "Tekken 7"
                                    :platform "my-platform"}})
```

It actually works and doesn't throw an error (as opposed to returning `String` for `Int` type).

Is it the correct behavior?

hlship:
It is true, we don't validate the data provided by the field resolver functions.  The question is whether we should and, if the value provided isn't a match, is that a soft error or a thrown exception?

itaied246:
Isn't that covered in graphql specs?
In `graphql-js` they return an error in the result.
Whatever you choose to implement, I think it is crucial to verify the enum, otherwise you open a door to broken API (think of database enums for example, no column can hold data other than the specified values).

PR #141, opened 2017-12-04
2017124 fix enum validation

itaied246:
#140 
Added failing tests on `enum` validation.


hlship:
Tests have failures. Also, I'm not certain what you are trying to fix. Please provide a more detailed title.

itaied246:
Hi, 
I've mentioned the bug in the issue.
I added a failing test to display the error, I don't know how to fix it though.

The bug is that an `enum` field is not enforced on a resolution.
I have created an object which returns a value not included in the enum values, and it works instead of throwing an error.

I have tested it on `graphql-js` and there the enum values are enforced and an error has been thrown.

PR #155, opened 2018-01-04
Fix null check for dynamic arguments

### LINKED ISSUE #133: Nullable input-object doesn't verify it contains all fields
This may be part of the graphql spec but I'm skeptical so I thought it's worth asking.

If we define a basic schema which has a mutation which takes an input object which is nullable, it's not enforced that all fields are passed within that nullable object.

```
:input-objects
{:create_good_habit_data
  {:description "The data required to create a new good habit."
   :fields
    {:name {:type (non-null String)}
     :description {:type String}
     :time_of_day {:type (non-null :habit_time)}
     :target_frequency {:type (non-null :create_frequency_data)}}}

:create_habit_data
  {:description "The data required to create a new habit."
   :fields
    {:good_habit {:type :create_good_habit_data}
     :bad_habit {:type :create_bad_habit_data}}}}}

:mutations
 {:add_habit
  {:type Int
   :description "Add a habit to your list of habits."
   :args {:create_habit_data {:type (non-null :create_habit_data)}}
   :resolve :query/resolve-mutation-add-habit}}
```

So as you can see above, we expect a good_habit/bad_habit or both or neither as they can be nulled. But if someone does create a good_habit (it's not `null`), I'd expect the fields to be enforced.

If for example we do this query:

```
mutation {
  add_habit(create_habit_data: {
    good_habit: {
      nonfield: "shouldn't be here" 
    }
  })
}
```

We will get an error (as expected) complaining that `nonfield` is an unexpected key. But if I were to pass an incomplete query: 

```
mutation {
  add_habit(create_habit_data: {good_habit: {name: "asdf"}})
}
```

No error is thrown, this is strange to me though as we are missing required fields on `good_habit` such as `time_of_day`.

Is this purposeful behavior to not check that all fields are passed for nullable input objects but do validation for unexpected keys and incorrect types on present keys? It seems it would be important to check all the non-null fields are present as well as that is what is specified in the schema.

Additionally, I've also tried`{:type (non-null :create_good_habit_data)}` and it still threw no error if I didn't pass all the fields, is this simply not being checked?

hlship:
Capture exception context at parse time for use with dynamic arguments
(from query variables)

Fixes #133

### DIFF
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -6,7 +6,7 @@
             [com.walmartlabs.lacinia.internal-utils
              :refer [cond-let update? q map-vals filter-vals
                      with-exception-context throw-exception to-message
-                     keepv as-keyword]]
+                     keepv as-keyword *exception-context*]]
             [com.walmartlabs.lacinia.parser.common :refer [antlr-parse parse-failures stringvalue->String]]
             [com.walmartlabs.lacinia.schema :as schema]
             [com.walmartlabs.lacinia.constants :as constants]
@@ -34,6 +34,11 @@
 
       :else
       (recur nested))))
+
+(defn ^:private non-null-kind?
+  "Peeks at the kind of the provided def (field, argument, or variable) to see if it is :non-null"
+  [any-def]
+  (-> any-def :type :kind (= :non-null)))
 
 ;; At some point, this will move to the schema when we work out how to do extensible
 ;; directives. A directive effector is invoked during the prepare phase to modify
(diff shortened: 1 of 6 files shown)


src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -325,7 +330,7 @@
 
     (let [object-fields (:fields schema-type)
           default-values (collect-default-values object-fields)
-          required-keys (keys (filter-vals :non-nullable? object-fields))
+          required-keys (keys (filter-vals non-null-kind? object-fields))
```
hlship: This is the heart of the bug.  Everything else was a bit drive-by for consistent error reporting when dynamic argument values are invalid (that is, consistent with reporting for static argument values).

PR #159, opened 2018-01-17
manually check compile args against spec

hiredman:
This change stops lacinia from turnning on spec assert checking globally
and manully checks the compile-args spec when compile is called.

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -1216,11 +1216,10 @@
   ([schema options]
    ;; This is based on clojure.spec's assert, but is always on
    ;; the single branch alt adds :args to the explain path as expected in tests
-   (when-not (s/valid? ::compile-args [schema options])
-     (let [ed (s/explain-data ::compile-args [schema options])]
-       (throw (ex-info
-               (str "Spec assertion failed\n" (with-out-str (s/explain-out ed)))
-               ed))))
+   (when-let [ed (s/explain-data ::compile-args [schema options])]
+     (throw (ex-info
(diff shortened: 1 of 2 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -1210,20 +1214,16 @@
   ([schema]
    (compile schema nil))
   ([schema options]
+   ;; This is based on clojure.spec's assert, but is always on
+   ;; the single branch alt adds :args to the explain path as expected in tests
+   (when-not (s/valid? ::compile-args [schema options])
+     (let [ed (s/explain-data ::compile-args [schema options])]
+       (throw (ex-info
+               (str "Spec assertion failed\n" (with-out-str (s/explain-out ed)))
```
hlship: Also, this looks like a bit of code that might be refactored out to internal-utils for re-use elsewhere.

hlship: How about `Arguments to compile do not conform to spec:`  ?

hiredman: looking at explain-data, it returns nil if the spec is invalid, so we don't need the extra valid? check, which simplifies things. I also added the custom exception message. Those two things together have me leaning towards not pulling it out as a utility, but let me know if you would still prefer that.

hlship:
It is necessary for you to sign the CLA (contributor license agreement) before we can merge.

Technically, you must sign the CLA before we can review the PR, but the CLA requirement is newer than the PR.

Please re-open when you are ready to sign the CLA.

hlship:
Please re-open this PR when you are ready to sign the CLA.  I'd love to merge this.

bcarrell:
Looks like some conflicts after I merged the other PR.  Mind handling?

PR #162, opened 2018-01-26
Add support for marking fields and enum values as deprecated

hlship:


src/com/walmartlabs/lacinia/introspection.clj
```diff
@@ -39,30 +40,48 @@
       (map #(type-name->schema-type schema %)
            interfaces))))
 
+(defn ^:private is-deprecated?
+  "The :deprecated key can either be a boolean, or a string which is the deprecation reason."
```
sashton: I could imagine the possibility of someone wanting to use a keyword as a deprecation value. However, I don't have a use-case for that providing additional value, so I think it's just something we could keep in mind, and not implement at this time.

hlship: You mean, the keyword would be interpreted as the name of a replacement field?

hlship: RE: includeDeprecated ... it's from the spec.  It defaults to false, which excludes deprecated elements.  Previously we ignored includeDeprecated because nothing could be deprecated.

hlship: includeDeprecated: it is ONLY for the introspection queries.  Outside of introspection, deprecated fields and enums behave exactly the same as other fields and enums.

PR #165, opened 2018-02-02
Be both more and less forgiving about enums

hlship:
- Failure to return a defined enum value (or nil) is now an exception, not a field error
- BUT it is legal again to return a string or symbol, which will be internally converted to a keyword


hlship:
I'll see about the conflicts ...

PR #166, opened 2018-02-12
Tutorial: "Initial Schema"; console output shows a call to `(start)` that is not needed, will not work

salmazov:
(start) is not working right here.

user=> (start)

CompilerException java.lang.RuntimeException: Unable to resolve symbol: start in this context, compiling:(/tmp/form-init7979298612564961133.clj:1:1)


hlship:
I'd like to merge this, but the CLA needs to be signed first.

hlship:
I'm looking into why it says the CLA status is pending.

PR #176, opened 2018-03-23
Make it possible to define operations as fields of explicit root objects

hlship:
Other GraphQL implementations always define operations in terms of root objects; we're coming into this from an odd angle.

In any case, this also supports a root object actually being a union, in which case we merge together all the fields of all the members.

This simplifies the SDL parsing logic, as it can shift a lot of work to schema/compile, which includes new checks that the SDL code did not make (the check for conflicting names).

The merging of fields purposely leaves the `:qualified-field-name` unchanged, so you might end up with `:__Queries/foo` mixed in with `:QueryRoot/bar` & etc. I think this means that any errors reported (or exceptions thrown) involving such operations will be easier to interpret.  Check out the new and changed tests for examples.

### DIFF
--- a/src/com/walmartlabs/lacinia/constants.clj
+++ b/src/com/walmartlabs/lacinia/constants.clj
@@ -1,14 +1,6 @@
 (ns ^:no-doc com.walmartlabs.lacinia.constants
   "A handy place to define namespaced constants")
 
-(def mutation-root
-  "Object in the compiled schema that contains, as fields, all mutations."
-  :MutationRoot)
-
-(def ^{:added "0.19.0"} subscription-root
-  "Object in the compiled schema that contains, as fields, all subscriptions."
-  :SubscriptionRoot)
-
 (def schema-key
   "Context key storing the compiled schema."
   ::schema)
--- a/src/com/walmartlabs/lacinia/constants.clj
+++ b/src/com/walmartlabs/lacinia/constants.clj
@@ -1,10 +1,6 @@
 (ns ^:no-doc com.walmartlabs.lacinia.constants
   "A handy place to define namespaced constants")
 
-(def query-root
-  "Object in the compiled schema that contains, as fields, all queries."
-  :QueryRoot)
-
 (def mutation-root
   "Object in the compiled schema that contains, as fields, all mutations."
   :MutationRoot)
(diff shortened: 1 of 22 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -257,10 +258,13 @@
...
                            ::graphql-identifier))
 (s/def ::graphql-identifier #(re-matches graphql-identifier (name %)))
-(s/def ::identifier (s/and (s/or :keyword simple-keyword?
-                                 :symbol simple-symbol?)
-                           graphql-identifier?))
-(s/def ::type (s/or :base-type ::identifier
+;; For style and/or historical reasons, type names can be a keyword or a symbol.
+;; The convention is that built-in types used a symbol, and application-defined types
+;; use a keyword.
+(s/def ::type-name (s/and (s/or :keyword simple-keyword?
```
hlship: I renamed ::identifier to ::type-name ... I got very tripped up thinking that every key in Lacinia could be keyword or symbol, when actually only type names are that special case.  Field names, argument names, etc. are restricted to simple keywords, and this name change makes that easier to understand at a glance.

src/com/walmartlabs/lacinia/constants.clj
```diff
@@ -1,10 +1,6 @@
 (ns ^:no-doc com.walmartlabs.lacinia.constants
   "A handy place to define namespaced constants")
 
-(def query-root
-  "Object in the compiled schema that contains, as fields, all queries."
-  :QueryRoot)
-
 (def mutation-root
```
hlship: I don't think this is needed anymore.

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -1131,10 +1131,29 @@
                (rest var-definitions)))))
 
 
-(def ^:private operation-type->root
-  {:query constants/query-root
-   :mutation constants/mutation-root
-   :subscription constants/subscription-root})
+;; Ultimately, this needs to be done once and cached, perhaps by the schema compiler itself.
+
+(defn ^:private operation-type->root
```
hlship: Ouch.  This seems to be a hangover form before I moved most of this logic into schema compilation.

hlship:
Just noticed a few things that were wrong from the diffs, I've just pushed new commits to address those.

PR #181, opened 2018-04-07
Wrap Antlr output in an easier to process intermediate format

hlship:
This is a big change: the parser namespace no longer invokes Anltr and operates on Antlr's output; an extra stage is in the middle that converts Antlr's output to a more easily worked with (and tested!) intermediate format.

Along the way, the line/column number locations have been improved and made more accurate.

Performance is unchanged, or perhaps minutely improved.

Many tests changed, but only because column and/or line numbers shifted: partly because we increment the column from Antlr (so that both line and column are 1-based) and partly because we can be smarter about what is identified as the location.  For example, for a `... on Foo`, the location will be on the `Foo`, not the `...`.


dev-resources/basic-vars.edn
```diff
@@ -1,7 +1,13 @@
-{:data {:default {:name "Darth Vader",
-                  :appears_in ["NEWHOPE" "EMPIRE" "JEDI"],
```
hlship: This change reflects that we haven't run the benchmarks in a while and correct results have changed (enum output is now always a keyword).

project.clj
```diff
@@ -12,6 +12,8 @@
                                   [expound "0.5.0"]
                                   [joda-time "2.9.9"]
                                   [com.walmartlabs/test-reporting "0.1.0"]
+                                  [io.aviso/logging "0.2.0"]
```
hlship: Having pedestal-log as a test/dev time dependency helped a lot!

test/com/walmartlabs/lacinia_test.clj
```diff
@@ -61,8 +61,8 @@
            (execute default-schema q {} nil {:operation-name "heroNameQuery"}))))
 
   (let [q "query heroNameQuery { hero { id name } } query dummyQuery { hero { id } }"]
-    (is (= {:errors [{:message "Multiple operations requested but operation-name not found.",
-                      :op-count 2,
+    (is (= {:errors [{:message "Multiple operations provided but no matching name found."
```
hlship: I reworked this error message, I think this version is more understandable.

src/com/walmartlabs/lacinia/parser/query.clj
```diff
@@ -0,0 +1,263 @@
...
+   :of-type (-> prod second xform)})
+
+(defmethod xform :typeName
+  [prod]
+  {:type :root-type
+   :type-name (-> prod second second xform)})
+
+(defmethod xform :selection
+  [prod]
+  (xform (-> prod second )))
```
thorstadt: I'd probably just do `(xform (second prod))` or `(-> prod second xform)`

hlship: I'll fix that at some point; not sure how it ended up like that.

PR #186, opened 2018-05-02
Tutorial: call to non-existing start function removed

matthiasn:
When following the tutorial, the start function does not exist yet, nor is it necessary to run the query.


matthiasn:
Thanks everyone for this library, this will be very helpful going forward.

PR #204, opened 2018-06-28
Allow variables inside nested fields of input objects

### LINKED ISSUE #189: Variables fails for object arguments
I receive an error similar to #164 when attempting this (valid) query:

```graphql
query test($a_string: String) {
  something(objParam: { objField: $a_string }) {
    id
  }
}

{ "a_string": "something" }
```

The returned error is:

```
Exception applying arguments to field `something': For argument `objParam', no method in multimethod 'process-dynamic-argument' for dispatch value: :object
```

Having the argument represent the entire object works, yet the exemplified query should work as well. A similar error happens with :array (ie: `objParam: [{ objField: $a_string }]`).

hlship:
Previously, a input object could be entirely in-line (in the GraphQL query) or entirely specified as a variable; this PR  allows
a mix where individual nested fields of an input object can be specified as variables.

This one gave me brain ache.

This builds on PR #203 

While I was there I made a minor improvement tPR #203 (properly reporting the exception so that extra meta data is identified) and made a tiny simplification/optimization.

Fixes #189


test/com/walmartlabs/input_objects_test.clj
```diff
@@ -55,3 +54,58 @@
...
                     {:g {:name "Backgammon"}}
                     nil)))))
+
+
+(deftest allows-for-variables-inside-nested-objects
+  (let [schema (compile-schema "input-object-schema.edn"
+                               {:queries/search (fn [_ args _]
+                                                  [(pr-str args)])})]
+    ;; First we make it easy, don't try to make it promote a single value to a list:
+    #_(is (= {:data {:search ["{:filter {:max_count 5, :terms [\"lego\"]}}"]}}
```
annapawlicka: Should this be commented out?

hlship: Missed that while I was fixing some issues; removing the `#_` and tests pass. Just pushed.

PR #211, opened 2018-07-12
Make error maps in the result conform to the June 2018 specification

hlship:
Error maps should only contain :message (a string), :locations,  :path, and :extensions.


src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -80,38 +80,6 @@
                            (cond-> node
                              (-> arguments :if false?) (assoc :disabled? true)))}}))
 
-(defn ^:private name-string-from
```
hlship: These should have been removed a release or two ago, they are no longer used.

test/com/walmartlabs/lacinia_test.clj
```diff
@@ -542,44 +539,9 @@
               :errors [{:message "Non-nullable field was null."
                         :locations [{:line 1
                                      :column 19}]
-                        :query-path [:hero :family :arch_enemy]}]}
+                        :path [:hero :family :arch_enemy]}]}
              executed)
-          "nulls the first nullable object after a non-nullable field returns null")))
-
-  ;; TODO tests below fail because of types not being expanded correctly
```
hlship: These tests have been commented out for a long time, at least a year.

PR #216, opened 2018-07-19
Rewrite the schema parser, the same as the query parser

hlship:
This is primarily refactoring to make it easier to support documentation strings in the schema language.

A few drive by fixes:

- The existing code parsed a default value as `:defaultValue`, but it should be `:default-value`
- The `:default-value` key for an field argument was not part of the spec

The new code also stops at the first key conflict it finds; the older code did a better job of exhaustively searching for those. Anything missed by the schema parser code is caught by `compile-schema`.

### DIFF
--- a/resources/com/walmartlabs/lacinia/schema.g4
+++ b/resources/com/walmartlabs/lacinia/schema.g4
@@ -27,19 +27,24 @@ subscriptionOperationDef
   ;
 
 typeDef
-  : 'type' typeName implementationDef? '{' fieldDef+ '}'
+  : 'type' typeName implementationDef? fieldDefs
   ;
+
+fieldDefs
+  : '{' fieldDef+ '}'
+  ;
+
 
 implementationDef
   : 'implements' typeName+
   ;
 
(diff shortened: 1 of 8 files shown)


resources/com/walmartlabs/lacinia/schema.g4
```diff
@@ -27,19 +27,24 @@ subscriptionOperationDef
   ;
 
 typeDef
-  : 'type' typeName implementationDef? '{' fieldDef+ '}'
+  : 'type' typeName implementationDef? fieldDefs
   ;
 
+fieldDefs
```
hlship: This made writing some parser code easier.

resources/com/walmartlabs/lacinia/schema.g4
```diff
@@ -98,6 +103,11 @@ defaultValue
   : '=' value
   ;
 
+BooleanValue
```
hlship: This needed to be moved up, before the `enumValue` rule; order matters in Antlr.

PR #219, opened 2018-08-09
Add (failing) test for duplicate Fragment names

eins78:
Currently lacinia does not validate uniqueness of Fragment names, so this test should be failing. 
I am sending it as a PR to gather feedback. I don't know if marking the test as "pending" and merging is an option (I know this from other testing tools). 

--- 

spec: <http://facebook.github.io/graphql/June2018/#sec-Fragment-Name-Uniqueness>
test in reference implementation: <https://github.com/graphql/graphql-js/blob/bce300f9db68a738dd30d2b8be81efb30b78690c/src/validation/__tests__/UniqueFragmentNames-test.js#L101-L132>

### DIFF
--- a/test/com/walmartlabs/lacinia/validator_test.clj
+++ b/test/com/walmartlabs/lacinia/validator_test.clj
@@ -191,6 +191,36 @@
                                    :line 9}]
                       :message "Fragment `HumanFragment' is never used."}]}
            (execute compiled-schema q {} nil))))
+  (let [q "query UseFragment {
+             luke: human(id: \"1000\") {
+               ...HumanFragment
+             }
+           }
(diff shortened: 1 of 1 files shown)


hlship:
Well, this is a nice start, but obviously can't be merged yet ... need a strategy for identifying the duplicate names.

I'm not familiar with the RI code, but I suspect that it does a bit more during query execution whereas Lacinia tries to do as much schema and query validation as possible in an earlier, compile-like, phase -- and some checks that are simpler at execution time are harder at compile time.

hlship:
This has been hanging around for a while and, and it is still on the back burner.  I'm closing it, but the PR can be referenced from an issue.

PR #231, opened 2018-09-28
Specify a version for clojure-future-spec

urzds:
Otherwise one would get following error: `Bad artifact coordinates clojure-future-spec:clojure-future-spec:jar:, expected format is <groupId>:<artifactId>[:<extension>[:<classifier>]]:<version>`


urzds:
CLA agreement will take a while.

PR #241, opened 2018-11-08
Convert the flatland.ordered dependency to vendored code

### LINKED ISSUE #237: JDK 11 Incompatibility
The dependency on `ordered` currently breaks JDK 11 compatibility (see amalloy/ordered#37). If/when `ordered` gets updated, updating the dependency would be the easiest fix, but the pull request has been open for 3 months.

hlship:
This also applies a trivial patch to the ordered set implementation,
to provide JDK 11 compatibility.

To keep any future maintenance of this code easy, I did nothing to the vendored code beyond changing the namespace.

Fixes #237

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,11 +3,16 @@
 Limited support for custom directives.
 
 **Breaking Change**: Previous releases of the Schema Definition Language parser
-wrapped the entire document with `{` and `}.
+wrapped the entire document with `{` and `}`.
 This is not correct, and such documents are not valid SDL.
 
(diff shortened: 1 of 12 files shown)


danielcompton:
Ordered has a new release [1.5.7](https://github.com/clj-commons/ordered/releases/tag/1.5.7) which contains a fix for this. If you'd like, you could revert this commit and upgrade to 1.5.7.

hlship:
At this point, unless they introduce a change we need, I think it's best to keep the code vendored.

PR #244, opened 2018-11-13
Add support for warnings and arbitrary changes to the :extensions map

hlship:
Previously, only Lacinia had the ability to write to the :extensions key (it used this for timings).

Warnings look like errors, but go in the [:extensions :warnings] key.

The extensions map can also be more arbitrary modified.

I also stumbled onto the fact that the benchmark data was broken, due to the previous fix related to nils vs empty arrays.


hlship:
Oh, I meant to submit documentation changes first.  Closing for a bit.

PR #245, opened 2018-11-16
Allow scalars to parse from any data, not just strings

hlship:
This allows a scalar to be a complex structure (validated by the scalar's parse function).  This ends up modeling scalars in a way very, very similar to how Clojure reader macros operate.

### DIFF
--- a/dev-resources/object-scalars-schema.edn
+++ b/dev-resources/object-scalars-schema.edn
@@ -1,0 +1,10 @@
+{:scalars
+ {:Data
+  {:description "Arbitrary JSON/EDN data."
+   :parse :scalars/data.parse
+   :serialize :scalars/data.serialize}}
+
+ :queries
+ {:echo {:type :Data
+         :args {:input {:type :Data}}
+         :resolve :queries/echo}}}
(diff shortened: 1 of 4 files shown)


solussd-reify:
awesome! I've wanted this!

candid82:
lgtm

hlship:
Looking at the RI, I believe you could implement a scalar in almost the exact same way, by providing a `parseLiteral` callback (in the RI, that gets the AST).

In the spec, the most problematic bit is this:

> All GraphQL scalars are representable as strings, though depending on the response format being used, there may be a more appropriate primitive for the given scalar type, and server should use those types when appropriate.

My interpretation of this, is that if you were implementing an UntypedMap scalar, `identity` wouldn't be enough: you should allow for either string or map, and (if string) you should parse the string as either JSON or EDN.

zk:
@hlship Sorry to resurrect this old issue, I was wondering if there's an appetite for adding the equivalent of UntypedArray scalars.

I was digging through the code and it seems that returned object properties that are considered sequential will result in a "Field resolver returned a collection of values, expected only a single value." even for scalar types.

Happy to attempt a patch here.

PR #246, opened 2018-11-16
Change the scalar transformers from conformers to simple callback functions

hlship:
This follows on from #245.

There are minor differences in how the scalar parse and serialize operations are implemented, and minor changes to how related error messages are generated (affecting tests).


hlship:
Going to change this such that a parser or serializer can return nil as a signal that the transformation failed, and will generate a generic error.

hlship:
The extra change, to allow `nil`, from a parse or callback, reflects a) how the RI handles those cases (though it uses `undefined`), and b) some ugly and redundant code in lacinia and in ereceipts-services related to these cases.  I think it's much better with nil being a shorthand for "could not parse/serialize".

Issue #249, opened 2018-11-20
Some introspection fields are now null when they should be empty lists


hlship:
PR #243 has revealed a problem in Introspection, where certain fields that must be emitted as a (possibly empty) list now are  emitted as nil, which causes client-side errors in the GraphiQL client.

This misbehavior was obscured by the old code path, which would silently convert many nils into an empty list.

We have to selectively force certain introspection related fields to return an empty list, rather than nil.

hlship:
A patch to the `__Type.interfaces` field seems to do the trick; the other affected fields would only be a problem if you, for example, define a union with no members, or an object with no fields, etc.

PR #251, opened 2018-11-27
Correct return empty lists (not nils) for certain introspection fields, in accordance with the specification

hlship:
Another change in this release, #243, had the unexpected consequence of making certain introspection related fields nil, when the prose of the spec identifies that they must be (empty) lists.

This had the effect of breaking GraphiQL, as it performed some client-side sanity checks that then failed.

This corrects those particular fields to get back into spec alignment.


hlship:
Fixes #249

PR #266, opened 2019-01-29
Allow keywords or symbols for modifiers, and * and ! as abbreviations

hlship:
This allows `'(! String)` rather than `'(non-null String)` for example, or `(:non-null :String)` (using keywords to avoid quoting), or even `[:! :String]`.

I've been thinking about this for a long time, especially every time I've typed `not-null` or `non-nil` instead of `non-null`.

*Wait for 0.33.0 to be active before merge*

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -38,8 +38,8 @@ In the schema, a type can be:
 
 - A keyword corresponding to an object, interface, enum, or union
 - A scalar type (built in, or schema defined)
-- A non-nillable version of any of the above: ``(non-null X)``
-- A list of any of the above: ``(list X)``
+- A non-nillable version of any of the above: ``(non-null X)`` [#non-null]_
+- A list of any of the above: ``(list X)`` [#list]_
 
 The built-in scalar types:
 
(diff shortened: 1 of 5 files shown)


solussd:
interesting– I keep reading `(! String)` as "not string". Any reason not to just support `<typename>!` ?

hlship:
Because we're a Lisp so we put the important part first?   I chose `!` as that's the non-null suffix in SDL, but perhaps something else would work better?

hlship:
Ah, I see what you mean; something like `String!` or even `String!*!` (instead of `(non-null (list (non-null String)))`.  I'm not sure I'm that comfortable with "interpreting" the type name symbol/keyword, even though that would be *very* concise.

solussd:
I don't have strong feelings about it, honestly. That said, the edn schemas are just data describing the schema and equivalent to SDL. I don't see `(non-null (list Thing))` as function application, but just data. (I haven't actually checked– maybe it really is being evaluated) `Thing!` is just data, with significance tied to the `!`. I understand not want to interpreting the symbol's meaning, other than it being a name referenced in a definition, though.

solussd:
I think I'd actually prefer the SDL syntax for type, e.g., `[Thing!]!`

hlship:
Maybe this entire approach is wrong, maybe type should also allow a string, not a symbol or keyword, and the string is parsed much like SDL.

I'm curious to see how @bcarrell weighs in.

hlship:
I'm leaning towards an alternative where you can use a string to specify the type as you would in a SDL document, eg. "[String!]" rather than `(list (non-null String))`.

PR #271, opened 2019-02-28
Do not allow an InputObject as a field type (outside of another IO)

### LINKED ISSUE #254: InputObject usage should be restricted to field arguments and other InputObjects, and forbidden in other places
From the #graphql slack, a user had a runtime issue inside GraphiQL that was traced back to a mutation returning an input object (rather than an object).

hlship:
Even some of our tests bent the rules here, and needed correction.

Fixes #254

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -1142,14 +1142,24 @@
   "Verifies that the type of every field and every field argument is valid."
   [schema object-def]
   (let [directive-defs (::directive-defs schema)
-        location (if (= :input-object (:category object-def))
+        input-object? (= :input-object (:category object-def))
+        location (if input-object?
                    :input-field-definition
                    :field-definition)]
     (doseq [field-def (-> object-def :fields vals)
             :let [field-type-name (extract-type-name (:type field-def))
-                  qualified-field-name (:qualified-name field-def)]]
-      (when-not (get schema field-type-name)
+                  qualified-field-name (:qualified-name field-def)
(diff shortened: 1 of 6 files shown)


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -1142,19 +1142,29 @@
...
+                  qualified-field-name (:qualified-name field-def)
+                  field-type (get schema field-type-name)]]
+      (when (nil? field-type)
         (throw (ex-info (format "Field %s references unknown type %s."
                                 (q qualified-field-name)
                                 (q field-type-name))
                         {:field-name qualified-field-name
                          :schema-types (type-map schema)})))
 
+      (when (and (not input-object?)
```
iku000888: @hlship Thanks for taking care of this! Was just running into similar issues. 

I believe the combo of input object containing a :object field type should be rejected as well.

The graphql spec also says input objects should not contain object types
https://facebook.github.io/graphql/June2018/#sec-Input-Objects

> The GraphQL Object type (ObjectTypeDefinition) defined above is inappropriate for re‐use here, because Object types can contain fields that define arguments or contain references to interfaces and unions, neither of which is appropriate for use as an input argument. For this reason, input objects have a separate type in the system.

hlship: Turns out that check is already present, but elsewhere. I'm moving the two checks closer together and making them report more similarly.

iku000888: Awsome! Thanks for the extra effort checking :bowing_man:

PR #277, opened 2019-04-05
Ensure that explicit null values are not converted to variable or argument defaults, as per latest spec

hlship:
This is inline, to the best of my understanding, with the intent, and the next revision of the spec.

Explicit null becomes a null.

Omitting a variable uses the default for the variable (if present), or the default for the argument (if present).


test/com/walmartlabs/lacinia/optimize_resolve.clj
```diff
@@ -1,77 +0,0 @@
-; Copyright (c) 2018-present Walmart, Inc.
```
hlship: This was inadvertently committed a ways back.  If you saw my tweet about naive vs. clever, this is the basis.

sashton:
@hlship This looks good to me, but I've got one additional related case: https://github.com/walmartlabs/lacinia/issues/278. Do you feel it fits in this PR or merits a separate one?

src/com/walmartlabs/lacinia/parser.clj
```diff
@@ -586,29 +585,31 @@
...
-            ;; variable has a default value that could be NULL
-            (contains? variable-def :default-value)
-            (:default-value variable-def)
+            ;; At this point, the argument is known to be nullable (if the argument is non-nullable,
+            ;; then the type check ensures that the variable is non-nullable as well)
+            ;; So if an explicit null was supplied we can just use that, because the argument
+            ;; has no default, and is nullable.
+
+            supplied?
+            nil
```
annapawlicka: I'm having trouble following this. Why do we return `nil` here? What if it's supplied and it's not a null value?

sashton: @annapawlicka My understanding is: if it had been a non-nil value, it would be handled earlier in this function: `(some? result)`

annapawlicka: Ooops, this diff hid didn't show it but I should have looked at the beginning of this fn myself. Thanks for pointing this out!

hlship:
Ok, I've added more logic for defaults on arguments and variables; I think we're in accordance with the June 2018 spec and beyond.

PR #281, opened 2019-04-26
Add option to control whether schema introspection is enabled

hlship:


chrisulloa:
Thank you! I had tested making `(introspection/introspection-schema)` with `with-redefs` return `nil` on our codebase, and it worked as expected fwiw.

PR #282, opened 2019-04-30
Refactor how wrapped values (e.g., with-error, etc.) are implemented

hlship:
This gets away from using a protocol and `reify` and instead there's just a single WrappedValue record and a case statement on what to do with wrapped values.  This simplifies the resolve namespace, and removes boilerplate code from each new kind of wrapper (the extra two methods defined by the old ResolverCommand protocol).

This builds on #281 which should be merged first.


hlship:
Darn, meant this to make it into 0.33.0.

PR #290, opened 2019-06-14
Improvements to the executor namespace preview API

hlship:
This makes a couple of enhancements to the preview API.

It fully supports using the preview API before query execution; the preview API then reflects the root operation(s).

A new function to get more useful data out of the selected fields.

Some benchmarks (see `perf`):

selections-tree:  134 µs
selections-tree + walk the tree (based on what eReceipts does): 479 µs
selections-seq: 78 µs
selections-seq2: 236 µs

The upshot is that we can get things done a little faster and without using undocumented/unsupported APIs.

### DIFF
--- a/test/com/walmartlabs/lacinia/selections_tests.clj
+++ b/test/com/walmartlabs/lacinia/selections_tests.clj
@@ -18,7 +18,6 @@
    [com.walmartlabs.lacinia.parser :as parser]
    [com.walmartlabs.lacinia.executor :as executor]
    [com.walmartlabs.test-utils :refer [compile-schema execute]]
-   [com.walmartlabs.lacinia.constants :as constants]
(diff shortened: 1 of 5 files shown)


test/com/walmartlabs/lacinia/selections_tests.clj
```diff
@@ -18,68 +18,105 @@
...
    (tree default-schema query))
   ([schema query]
    (->> query
-        (app-context schema)
+        (parse-and-wrap schema)
         executor/selections-tree)))
 
 (deftest simple-cases
-  (is (= [:character/name]
+  (is (= [:__Queries/hero :character/name]
```
hlship: Lots of changes; since we generally parse from the query root, the root operations are now visible and need to be accounted for in the test results.

PR #295, opened 2019-08-14
Allow a var, as well as a function, for streamers and default resolver to support REPL-oriented development

olimsaidov:
Relates to #292


hlship:
Thanks for addressing this oversight!

PR #301, opened 2019-10-15
Support timeouts when executing query

hlship:
This involved moving much of the execution of the query into a secondary thread, either via `future` or within the configured callback executor.


hiredman:
It isn't clear to me what the race fixed in 
3646657 is. If it is that when a callback and a result are present and on-deliver! is called, the new callback is invoked instead of getting an error, I think that could be fixed by re-ordering the cond clauses without introducing the lock. Or if you want to introduce the lock you could get rid of the atom by using clojure.lang.Box or a deftype with mutable fields. Using both the lock and an atom seems distasteful.

hlship:
Yes, I thought there was a race condition, it was actually a problem in my application's test code, a terminated thread pool executor.  I then switched back to compare-and-set! because it was faster and, once you get used to it, easier to understand.

src/com/walmartlabs/lacinia.clj
```diff
@@ -57,15 +57,30 @@
...
+          :or {timeout-ms 0
+               timeout-error {:message "Query execution timed out."}}} options
+         execution-result (execute-parsed-query-async parsed-query variables context)
+         result (do
+                  (resolve/on-deliver! execution-result *result)
+                  ;; Block on that deliver, then return the final result.
+                  (if (pos? timeout-ms)
+                    (deref *result
+                           timeout-ms
+                           {:errors [timeout-error]})
```
escherize: Why not use a namespaced keyword to indicate an error here?

They're:
1) less likely to collide with `@*result`'s return value
2) more explicit
3) it's easier to grep the code to figure out what went wrong

hlship: That's a GraphQL error result.  So, when it times out, we provide a proper GraphQL result.  There isn't a good way to terminate the ongoing processing inside the new thread, but the GC should clean it up.

hlship: Maybe we're not talking about the same thing.  

I guess we have a tradition of returning a _value_ on an _exception_.  But that's mostly in the eReceipts code, not something we have to deal with in Lacinia normally.

src/com/walmartlabs/lacinia/resolve.clj
```diff
@@ -249,7 +266,7 @@
                                 ;; of wrapped values.
                                 (recur (cons value wrapped-values)
                                        (:value value))
-                                (let [new-value (wrapper-fn context args value value)]
+                                (let [new-value (wrapper-fn context args initial-value value)]
```
hlship: This is a drive-by fix.

escherize:
lgtm

PR #302, opened 2019-10-18
When invoking a promise callback, stay on same thread if already a pooled thread

hlship:
This builds on #301 which should be merged first.

There is a small performance improvement here ... I think. The schemas in the benchmarks don't make heavy use of resolver result promises, though some of the internals of Lacinia do.

Essentially, the idea is that if already inside a pooled thread to perform a callback, just invoke the callback directly.  In a lot of cases, resolving one promise causes a cascade of promise resolution, and this eliminates the cost of obtaining a pooled thread.  

However, I'm concerned that this will interfere with concurrency, so I'm checking into that as well.


hlship:
[exec-time.pdf](https://github.com/walmartlabs/lacinia/files/3758630/exec-time.pdf)

Can see a modest performance improvement via this optimization.

hlship:
Just realized there's an issue with thread bindings that may not be quite right with the way we're managing bindings.  Need to check that the code is right.

src/com/walmartlabs/lacinia/resolve.clj
```diff
@@ -166,9 +169,12 @@
 
             (if (compare-and-set! *state state (assoc state :resolved-value resolved-value))
               (when-let [callback (:callback state)]
-                (if-some [^Executor executor *callback-executor*]
-                  (.execute executor (bound-fn [] (callback resolved-value)))
```
hlship: This was never right; there's no reason to execute the callback with per-thread bindings when the promise is resolved; the client code that sets the callback should be responsible for using bound-fn if it needs access to the per-thread bindings.

PR #303, opened 2019-10-23
Update custom-scalars.rst

miikka:
As far as I can tell, the Int parser does not handle strings anymore.


miikka:
Actually I don't really want to think through the CLA so let's not do this.

hlship:
I'm sorry you feel that way.  CLAs are necessary because otherwise the rights to an open source project become murky. I don't feel the need to sign the CLA should be an imposition.

PR #306, opened 2019-10-28
Improve parallelism

hlship:
The current implementation of aggregating ResolveResult[Promise]s is inefficient when the RRs are already resolved (the common case) and prevent later work from happening when earlier work is hidden behind an un-resolved RRP.

This code optimizes that case, allowing Lacinia to process more of the query across multiple threads in parallel.


hlship:
There's some issue with the CircleCI build.  Not sure why it fails there and works locally.

PR #308, opened 2019-11-08
Add :parse and :serialize to Enum, similar to what exists for Scalar

### LINKED ISSUE #221: Allow application to transform enum values between GraphQL values an application-specified values
It would be nice to have enum resolvers, or maybe enum transformations (e.g., parser/serializer) to allow for internal representations of enum values as namespaced keywords. For example, if I had an enum type `Role` with possible values `ADMIN`, `MEMBER`, and `GUEST` that I want to represent internally as `:role/admin`, `:role/member`, and `:role/guest` I could have a transformer that manages that transformation.

Currently I have to perform the mapping manually or heuristically.

There is some prior art in Apollo: https://www.apollographql.com/docs/graphql-tools/scalars.html#internal-values

I'll add: I'm happy to open a PR for this if it makes sense.

hlship:
Fixes #221


src/com/walmartlabs/lacinia/schema.clj
```diff
@@ -690,22 +695,28 @@
...
+                   (let [possible-values (-> field-type :values set)
+                         serializer (:serialize field-type)]
                      (fn validate-enum [{:keys [resolved-value]
                                          :as selector-context}]
                        (cond-let
+                         ;; The resolver function can return a value that makes sense from
+                         ;; the application's model (for example, a namespaced keyword or even a string)
+                         ;; and the enum's serializer converts that to a keyword, which is then
+                         ;; validated to match a known value for the enum.
+                         :let [serialized (serializer resolved-value)]
```
miikka: Just noticed: because `serializer` is `as-keyword` by default, you can't return `nil` as an enum value for nullable enum anymore.

hlship: I hadn't thought of that.  Should be a `(when resolved-value ...)` then.

miikka: Alternatively this `:let` could be moved after the next `(nil? resolved-value)` block

hlship: Moved the let, added tests.

PR #309, opened 2019-11-26
"Undefined enum value" errors return selector errors instead of throwing

gusbicalho:
# Current behavior:
After running a resolver, if we validate an enum field and find an invalid value, we throw an exception.

This seems sensible on the surface. However, if this happens while running an asynchronous resolver, the exception is throw in the secondary thread, after it `deliver!`s the promise. At that point, we cannot do anything with this error: the promise has already been delivered, so we cannot deliver a `(with-error ...)` value.

The result is that the secondary thread swallows the exception, and the main thread hangs forever, because we never finished handling the callback.

# Solution
When we find an undefined enum value, we return a `selector-error` instead of throwing an exception, just like we do when we have a coercion failure, for example.

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -711,10 +711,10 @@
                          :let [serialized (serializer resolved-value)]
 
                          (not (possible-values serialized))
-                         (throw (ex-info "Field resolver returned an undefined enum value."
-                                         {:resolved-value resolved-value
-                                          :serialized-value serialized
-                                          :enum-values possible-values}))
+                         (selector-error selector-context (error "Field resolver returned an undefined enum value."
+                                                                 {:resolved-value resolved-value
+                                                                  :serialized-value serialized
+                                                                  :enum-values possible-values}))
 
                          :else
                          (selector (assoc selector-context :resolved-value serialized)))))
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -660,7 +660,7 @@
 
         selector (if (= :scalar category)
                    (let [serializer (:serialize field-type)]
-                     (fn select-coerion [selector-context]
+                     (fn select-coercion [selector-context]
                        (cond-let
 
                          :let [{:keys [resolved-value]} selector-context]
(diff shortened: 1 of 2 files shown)


hlship:
I'm on the fence on this; if there's a problem with your schema, you should be seeing server-side exceptions, not client side errors.  How will your monitoring alert you to this problem in your schema?  Why are such things not caught in your tests.

I have previously found that catching exceptions thrown by field resolver functions actually made it much harder to identify problems.  This is reverting towards that mistake.

gusbicalho:
Hi @hlship , thanks for looking into this!

This is not an exception thrown by a field resolver, it's an exception thrown by a validation inside lacinia. There is no way (that I'm aware of) to handle it and wrap it in `with-error` to be able to return a valid GraphQL response.

However, the problem is not the exception itself. And It would not be that bad if the exception bubbled up and ending up with a status 500, even though this is not what a GraphQL endpoint should do.

The problem is that, if this happens in an asynchronous resolver, it leads to the request thread hanging forever. It never returns, and holds a thread in the server. In fact, this was what happened in my tests - instead of failing, they hang!

Besides, if I had an exception bubbling up the main thread leading to a 500, I could monitor and build an alarm for that. With a hanging thread I'd probably just get a server that eats more and more memory until it crashes mysteriously.

So you're right in that my tests should cover this, but I think it's a good idea to have a more sane error behaviour. It's expected that some bad code could lead to exceptions, but having a request hanging forever is really surprising.

hlship:
I think there is a bigger case, that any exception in code invoked by lacinia should be caught, captured, and be re-thrown at the top-level.  So I'll think about that; basically another atom at the top level that can dismantle the entire execution.

hlship:
I'd like to look at your solution, but there are now conflicts to be resolved.

gusbicalho:
@hlship I just fixed the conflicts. Can you take a look?

hlship:
Thanks!  I'll look at it shortly.

hlship:
Not forgotten about ... just been busy with work and other things.  Want to review this and get out a RC I think.

hlship:
I've been thinking about this hard for a few days.  On the one hand, having a resolver return a enum value that can't be serialized to the client is a programmer's bug and should be handled as such BUT the behavior provided in this PR matches what we currently do for other scalar types.

Ultimately, my thought was ... if a programmer deploys a change with unintended consequences (oops, returned a value not mapped in my enum) then having the entire request fail with a 500 is somewhat against the GraphQL philosophy of returning as much data as possible.

I think we want to investigate a logging option in the future, since I'd like to accompany this kind of failure (as well as scalar serializer faliures) with a loud ERROR to get the attention of the developer, even as the deployed app limps along.

PR #323, opened 2020-06-03
Improve exceptions by identifying field and other data

hlship:
This PR ensures that uncaught exceptions thrown by resolver functions, or by code immediately after (such as field serializers, etc.) are caught and re-wrapped with more information (the qualified field name, the query location and path, the field arguments, and so forth).

### DIFF
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -1,30 +1,30 @@
-; Copyright (c) 2017-present Walmart, Inc.
-;
-; Licensed under the Apache License, Version 2.0 (the "License")
-; you may not use this file except in compliance with the License.
-; You may obtain a copy of the License at
-;
-;     http://www.apache.org/licenses/LICENSE-2.0
-;
-; Unless required by applicable law or agreed to in writing, software
-; distributed under the License is distributed on an "AS IS" BASIS,
-; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-; See the License for the specific language governing permissions and
-; limitations under the License.
+;; Copyright (c) 2017-present Walmart, Inc.
+;;
+;; Licensed under the Apache License, Version 2.0 (the "License")
+;; you may not use this file except in compliance with the License.
+;; You may obtain a copy of the License at
+;;
+;;     http://www.apache.org/licenses/LICENSE-2.0
+;;
+;; Unless required by applicable law or agreed to in writing, software
+;; distributed under the License is distributed on an "AS IS" BASIS,
+;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;; See the License for the specific language governing permissions and
+;; limitations under the License.
 
 (ns com.walmartlabs.lacinia.executor
   "Mechanisms for executing parsed queries against compiled schemas."
   (:require
-    [com.walmartlabs.lacinia.internal-utils
-     :refer [cond-let map-vals remove-vals q aggregate-results transform-result to-message]]
-    [flatland.ordered.map :refer [ordered-map]]
-    [com.walmartlabs.lacinia.schema :as schema]
-    [com.walmartlabs.lacinia.resolve :as resolve
-     :refer [resolve-as resolve-promise]]
-    [com.walmartlabs.lacinia.selector-context :as sc]
-    [com.walmartlabs.lacinia.constants :as constants])
+   [com.walmartlabs.lacinia.internal-utils
+    :refer [cond-let map-vals remove-vals q aggregate-results transform-result to-message]]
+   [flatland.ordered.map :refer [ordered-map]]
+   [com.walmartlabs.lacinia.schema :as schema]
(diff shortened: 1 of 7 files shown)


src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -1,30 +1,30 @@
...
   (:require
-    [com.walmartlabs.lacinia.internal-utils
-     :refer [cond-let map-vals remove-vals q aggregate-results transform-result]]
-    [flatland.ordered.map :refer [ordered-map]]
-    [com.walmartlabs.lacinia.schema :as schema]
-    [com.walmartlabs.lacinia.resolve :as resolve
-     :refer [resolve-as resolve-promise]]
-    [com.walmartlabs.lacinia.selector-context :as sc]
-    [com.walmartlabs.lacinia.constants :as constants])
+   [com.walmartlabs.lacinia.internal-utils
```
hlship, line 18: I ran our indenter on the code, so there's a lot of whitespace changes.

test/com/walmartlabs/lacinia/enums_test.clj
```diff
@@ -94,15 +92,6 @@
                                    :serialized-value :ok}}]}
            (utils/execute schema "{ current_status }")))))
 
-(deftest enum-resolver-must-return-named-value
```
hlship, line 97: Essentially, this test has moved over the resolver errors test.

test/com/walmartlabs/lacinia/enums_test.clj
```diff
@@ -183,12 +172,12 @@
 
     (is (= {:data {:echo {:input ":fu/bar"
                           :output :bad}}}
-           (-> (lacinia/execute schema
-                                "query ($in : status) {
+           (-> (execute schema
```
hlship, line 175: I also ran the linter and found a couple of things.

test/com/walmartlabs/lacinia/resolver_errors_test.clj
```diff
@@ -1,39 +1,44 @@
...
+   [com.walmartlabs.lacinia.resolve :refer [resolve-as]]
+   [com.walmartlabs.test-utils :refer [execute] :as utils]
+   [com.walmartlabs.lacinia.schema :as schema])
+  (:import (clojure.lang ExceptionInfo)
+           (java.awt Color)))
 
-(def ^:private failure-exception (ex-info "Fail!" {:reason :testing}))
+(defn ^:private resolve-exception
+  [_ _ _]
+  (throw (ex-info "Fail!" {:reason :testing})))
```
hlship, line 27: This was changed while I was making sure that catching/rethrowing the exception didn't hide or obfuscate the original stack trace into the resolver function.

chrisulloa:
This is actually really helpful for our use case. Recently I had changed the majority of our resolvers to async resolvers, so that we could manage multiple resolver network calls in parallel, but noticed that uncaught errors for async resolvers were really difficult to diagnose.

I ended up writing a wrapper for all of our async resolvers to make sure exceptions are caught:
```
(defmacro async-resolver
  [& resolver-body]
  `(let [p# (resolve-promise)]
     (future
      (try
        (let [result# (do ~@resolver-body)] <logic to deliver! result/errors>)
        (catch Exception e#
          (log/warnf "Uncaught error in async resolver: %s" (.getMessage e#))
          (deliver! p# nil {:message (.getMessage e#)}))))
     p#))
```
Because our consumers make large queries, we still have to make sure all our resolvers wrap exceptions and return information about where the error occurred. Otherwise we would sometimes end up with cryptic log messages like "Uncaught error in async resolver: nil" (based on code above).

Having Lacinia return all that metadata about where the exception occurred would be really valuable. I'd be happy to give it a go and give further feedback.

src/com/walmartlabs/lacinia/executor.clj
```diff
@@ -121,52 +121,65 @@
...
+      (let [field-name (get-in field-selection [:field-definition :qualified-name])
+            {:keys [location arguments]} field-selection]
+        (throw (ex-info (str "Exception in resolver for "
+                             (q field-name)
+                             ": "
+                             (to-message t))
+                        {:field-name field-name
+                         :arguments arguments
+                         :location location
+                         :path (:path execution-context)}
```
chrisulloa, line 167: This is all really helpful info

hlship:
Of course, as soon as you put async into the mix, an exception thrown by resolver isn't and can't be caught by Lacinia.  You tend to see this with core.async and go blocks, where you can occasionally get an exception that doesn't have anything but JVM and core.async stack frames (no lacinia, no application code).  Those can be very fun to figure out!

mmmdreg:
@hlship interested to know what the rationale was for removing exception handling back in 0.19.0 and then adding them again? In any case this is a good change, especially for the principle of least surprise. 

From the 0.19.0 change log:

> Lacinia no longer catches and reports exceptions inside field resolvers.

hlship:
The older code remove at 19 was catching the exception and then converting it into errors on the field.  The result of that, especially inside async code paths, was it was really hard to know what went wrong and where.

The new change captures the exception inside the field resolver, wraps it in a new exception to provide useful context, and rethrows it.

PR #326, opened 2020-06-12
Add non-empty-list support

CHNB128:


hlship:
We will not be going forward on this, as it can't be properly supported in introspection, or in federation (something I'm working on).

PR #331, opened 2020-07-24
Use possible enum serialisation in introspection of default value

lennartbuit:
With custom scalars, if you have a default value in your schema.edn, you can use a 'parsed' value and for introspection, lacinia will transform it back to a 'serialized' value.

However, with enums, you can have a 'parsed' value in your schema, but there was no logic to convert it back to a serialized value prior to introspection. That left us between a rock and hard place, we could put a non-parsed default value in the schema, but then the enums parse function would not run prior to passing it as an argument (but introspection was correct), or we could put a parsed default value in the schema, but then the introspection result would be wrong.

Like scalars, this change invokes a possibly customly defined serialize function prior to returning a default value for introspection. So, if you put a parsed enum value in schema.edn, all is well with introspection.

### DIFF
--- a/src/com/walmartlabs/lacinia/introspection.clj
+++ b/src/com/walmartlabs/lacinia/introspection.clj
@@ -241,8 +241,10 @@
       (str serialized))))
 
 (defmethod emit-default-value :enum
-  [_ _ value]
-  (name value))
+  [schema type-map value]
+  (let [type-name  (:type type-map)
+        enum-def   (get schema type-name)]
+    (name ((:serialize enum-def) value))))
 
 (defmethod emit-default-value :list
   [schema type-map value]
--- a/test/com/walmartlabs/introspection_test.clj
+++ b/test/com/walmartlabs/introspection_test.clj
@@ -13,9 +13,13 @@
 ; limitations under the License.
 
 (ns com.walmartlabs.introspection-test
-  (:require [clojure.test :refer [deftest testing is]]
+  (:require [clojure.edn :as edn]
+            [clojure.java.io :as io]
+            [clojure.string :as str]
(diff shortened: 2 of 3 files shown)


test/com/walmartlabs/introspection_test.clj
```diff
@@ -1472,6 +1476,26 @@
              __type(name: \"Filter\") { inputFields { name defaultValue } }
            }")))))
 
+(deftest enum-transformer-default-value
+  (let [;; Create kebab-cased namespaced keywords for enum values
+        parse-country     (fn [code] (keyword :country-code (str/lower-case (name code))))
+        ;; upper case, non namespaced keywords are in the schema
+        serialize-country (fn [code] (str/upper-case (name code)))
+        schema (-> (io/resource "enum-default-value-with-transformer-schema.edn")
```
lennartbuit, line 1484: Not particularly happy about this test clutter, but the `utils/compile-schema` function is not capable of adding enum transformers. Taking suggestions :)!

src/com/walmartlabs/lacinia/introspection.clj
```diff
@@ -241,8 +241,10 @@
       (str serialized))))
 
 (defmethod emit-default-value :enum
-  [_ _ value]
-  (name value))
+  [schema type-map value]
+  (let [type-name  (:type type-map)
+        enum-def   (get schema type-name)]
+    (name ((:serialize enum-def) value))))
```
lennartbuit, line 247: There is a default serialize function of `as-keyword`. So; the assumption here is that a consumers `serialize` function returns a string (in which case `name` is identity), or a keyword (in which case `name` strips the colon).

hlship:
Thanks for this, a very necessary improvement.

PR #347, opened 2020-11-13
mutable-context.edn typo: Requires lacina instead of lacinia

jkjuopperi:
```
[com.walmartlabs.lacina.resolve :as resolve]
```


hlship (approved):

hlship:
Thanks for the catch and the fix!

PR #363, opened 2021-04-10
Ensure fragments aren't cyclic

### LINKED ISSUE #321: Infinite loops in query
I'm wondering if figuring out infinite loops should be a concern of Lacinia's. It seems so to me, given that it's a kind of query error.

Here's an example using the [GitHub API](https://developer.github.com/v4/explorer/):

```graphql
{
  repository(name: "eros", owner: "xiongtx") {
    ...RepoFields
  }
}

fragment RepoFields on Repository {
  assignableUsers(first: 5) {
    edges {
      node {
        ...UserFields
      }
    }
  }
}

fragment UserFields on User {
  repositories(first: 5) {
    edges {
      node {
        ...RepoFields
      }
    }
  }
}
```

GitHub responds w/:

```json
{
  "errors": [
    {
      "path": [
        "fragment RepoFields"
      ],
      "extensions": {
        "code": "infiniteLoop",
        "fragmentName": "RepoFields"
      },
      "locations": [
        {
          "line": 7,
          "column": 1
        }
      ],
      "message": "Fragment RepoFields contains an infinite loop"
    },
    {
      "path": [
        "fragment UserFields"
      ],
      "extensions": {
        "code": "infiniteLoop",
        "fragmentName": "UserFields"
      },
      "locations": [
        {
          "line": 17,
          "column": 1
        }
      ],
      "message": "Fragment UserFields contains an infinite loop"
    }
  ]
}
```

hlship:
Fixes #321

### DIFF
--- a/src/com/walmartlabs/lacinia/internal_utils.clj
+++ b/src/com/walmartlabs/lacinia/internal_utils.clj
@@ -436,3 +436,12 @@
                            (fn [value]
                              (resolve/deliver! xformed (xf value))))
       xformed)))
+
+(defn seek
+  "Returns the first value of coll for which pred returns a truthy value."
+  [pred coll]
+  (reduce (fn [_ v]
+            (when (pred v)
+              (reduced v)))
+           nil
+          coll))
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@ Some additions and incompatible changes have been made to the protocols in the `
 these protocols are not meant to be implemented except by the library, and prior code (from 0.38) will be
 largely source compatible with the new names.  Further such changes are expected.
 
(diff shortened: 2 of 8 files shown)


src/com/walmartlabs/lacinia/internal_utils.clj
```diff
@@ -436,3 +436,12 @@
                            (fn [value]
                              (resolve/deliver! xformed (xf value))))
       xformed)))
+
+(defn seek
```
oliyh, line 440: Isn't this the same as `some`?

hlship: No, seek is more like filter -> first, in that it returns the value for which the predicate returns true, not the predicate itself.

PR #366, opened 2021-07-01
Support non-null object fields that implement nullable interface fields.

GreshamDanielStephens:
Hi there,
Thanks for the very useful library!

I found this bug while trying to write some graphql of the form:
```
interface IA {
    a: String
}

type A implements IA {
    a: String!
}
```

I read through https://spec.graphql.org/June2018/#ObjectTypeDefinition and from my understanding of:
> 4. An object field type is a valid sub‐type if it is a Non‐Null variant of a valid sub‐type of the interface field type.

is that it is intended to be supported.

Looking through the existing code it looks like Lacinia also intended to support it, but had a bug.
Hopefully this is a reasonable fix to the issue.

I'm not 100% sure whether I've got the tests in the right place, or if I could do something better than `is some?`. Please let me know if there is a better alternative.

All the best,
Daniel

### DIFF
--- a/src/com/walmartlabs/lacinia/schema.clj
+++ b/src/com/walmartlabs/lacinia/schema.clj
@@ -663,7 +663,7 @@
       ;; the object can be more specific than the interface.
       (and (= o-kind :non-null)
            (not= i-kind :non-null))
-      (recur schema i-kind o-type)
+      (recur schema interface-type o-type)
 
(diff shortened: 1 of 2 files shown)


hlship:
Thanks; I'm looking this over.

hlship (approved):

GreshamDanielStephens:
Thanks for the quick response! Hope you have a nice day :wave:

PR #375, opened 2021-08-23
Add BigInt/BigInteger/BigDecimal support to int and float serialization

seeday:


seeday:
I've signed the CLA, thanks for your patience.

hlship (approved):

PR #376, opened 2021-09-04
Update init-schema.rst

hagenek:
typofix


hlship:
It's correct as stands; we'll be talking about the GraphiQL IDE (provided by lacinia-pedestal).  Thanks for contributing though!

PR #384, opened 2021-10-29
Performance improvements

hlship:
We've found that our services are spending too much time executing Lacinia queries, even after all I/O has completed (we do all of our I/O in a root resolver); this PR makes a number of changes to reduce the amount of ephemeral objects created during execution; we get away from using a single "whiteboard" map (the execution context) and instead, use ordinary function parameters.  It makes for ugly, but more efficient code.

These changes result in about a 30% improvement in our largest benchmark, the "deep" benchmark (from 11.48ms to 7.40ms).


test/com/walmartlabs/lacinia/custom_scalars_test.clj
```diff
@@ -386,7 +386,7 @@
               :errors [{:message "Non-nullable field was null.",
                         :locations [{:column 31
                                      :line 2}]
-                        :path [:shout]
+                        :path [:shout 0 0 0]
```
hlship, line 389: This was a latent bug fixed in the new implementation.

PR #390, opened 2021-11-05
Update fields.rst

yyna:
typo


hlship:
Thanks for the fix!  Even something as minor as this needs the CLA signed, I'm afraid.

yyna:
I already signed up, check this PR - https://github.com/walmartlabs/lacinia/pull/372

PR #392, opened 2021-11-15
fix `resolve-as` behaviour in subscriptions

thumbnail:
While testing out the latest performance improvements, and noticed that the global `:error` key wasn't working for subscriptions specifically. 27f251b00358197f7a75a7b8b4ba3e3671d5214a adds a test asserting the expected behaviour. Using `git bisect` I noticed the behaviour was happening since f6371f3bf3fc93d9f3623b6352bf0933bf198dc7.

As a 'quick fix' I reverted the change with f4a95981f6bf41e50e08f93cd02b8db3b25ff850 to fix the test; definitely open for suggestions though :)

### DIFF
--- a/dev-resources/subscriptions-schema.edn
+++ b/dev-resources/subscriptions-schema.edn
@@ -7,5 +7,7 @@
  :subscriptions
  {:logs
   {:type :LogEvent
-   :args {:severity {:type String}}
+   :args {:severity {:type String}
+          :fakeError {:type Boolean
+                      :default false}}
    :stream :stream-logs}}}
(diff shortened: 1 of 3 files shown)


hlship:
Reverting the change is not going to happen, since that was one of the key performance improvements I'm committed do.

At issue here is the use of `resolve-as` to identify an error associated with the value provided to the stream function; this isn't something initially envisioned, as the stream function is supposed to be invoked with _values_ and we've tightened up the definition of value to generally not include a ResolverResult (but to include a WrappedValue).

The other issue is how, due to evolution of the APIs and contracts, resolve-as has two very different purposes: one to ensure that a resolver function returns a ResolverResult, and secondarily, to associate an error with a true value. This can not longer be separated, which is unfortunate.

So to address both these needs, I think this should be handled in between the stream function and the resolve as a special case of unwrapping a ResolverResult and unwrapping WrappedValues before passing those to the resolver.

hlship:
I'll put something together after lunch.

hlship:
Part of this fix needs to be in lacinia-pedestal.

PR #402, opened 2022-02-11
Add explicit dependency on io.aviso/pretty

robhanlon22:
https://github.com/walmartlabs/lacinia/blob/master/src/com/walmartlabs/lacinia/trace.clj requires io.aviso.exception, which is transitively pulled in through io.aviso/logging, even though it is an explicit project dependency.


hlship:
This is a duplicate of #408.

PR #403, opened 2022-02-11
trace namespace depends on dev-only dependencies

robhanlon22:
I attempted to upgrade to Lacinia 1.1, but production builds do not work because the trace namespace depends on a dev-only dependency (io.aviso/pretty through io.aviso/logging), and the no_unused_fragments pulls in the trace namespace. As a quick fix, remove the unused require; going forward, it looks like io.aviso needs to be included as a non-dev dep. Not sure if this is the path that should be taken, so I am refraining from adding it.


hlship:
My plan is to move the trace code to its own library, since I find it useful and want to use it on other projects, some that do not have a Lacinia dependency.

hlship:
This is a duplicate of #408.

PR #404, opened 2022-02-15
Fix typo: deprecation -> deprecated

yyna:


hlship:
Even for something as trivial as this (I presume, I can't look) I need to see a signed CLA.  Thanks in advance!

yyna:
Check this. :)
https://github.com/walmartlabs/lacinia/pull/372

hlship (approved):

PR #420, opened 2022-08-08
Modify federation to be possible with edn schema as well

hmy3743:
Extends lacinia's support for federation to add the ability to support federation through the edn schema as well.


deps.edn
```diff
@@ -1,7 +1,8 @@
 {:deps {org.clojure/clojure {:mvn/version "1.10.3"}
         clj-antlr/clj-antlr {:mvn/version "0.2.12"}
         org.flatland/ordered {:mvn/version "1.15.10"}
-        org.clojure/data.json {:mvn/version "2.4.0"}}
+        org.clojure/data.json {:mvn/version "2.4.0"}
+        org.clojure/core.match {:mvn/version "1.0.0"}}
```
hlship, line 5: Looking at how `match` is used in the code, it feels a bit like over kill - most of the cases could be covered with `cond` or `condp`, and adding dependencies is always an issue.

dev-resources/edn-federation.sdl
```diff
@@ -0,0 +1,34 @@
+type _Service{
```
hlship, line 1: I wonder if we could get pretty-printed output. It would make things more complicated, for sure.

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +121,215 @@
...
+  [f x]
+  (if (-> x first seq?)
+    (apply f x)
+    (f x)))
+
+(defn ^:private edn-description->sdl-description
+  [description]
+  (if (nil? description)
+    ""
+    (str "\"\"\"\n" description "\n\"\"\"\n")))
```
hlship, line 134: What if description itself includes characters, such as `"`, that need to be escaped?

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +121,215 @@
...
+        'Boolean "Boolean"
+        'ID "ID"
+        (object :guard keyword?) (name object)
+        (scalar :guard symbol?) (name scalar)))
+    (recur (list type))))
+
+(defn ^:private value->string
+  [value]
+  (match value
+    (string :guard string?) (str "\"" string "\"")
```
hlship, line 156: Same here, may need to escape some characters.

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +121,215 @@
...
+  [objects]
+  (->> objects
+       (map (fn [[key {:keys [fields directives description]}]]
+              (str (edn-description->sdl-description description)
+                   "type "
+                   (name key)
+                   (edn-directives->sdl-directives directives)
+                   (edn-fields->sdl-fields fields))))
+       (join "\n")))
+(defn ^:private edn-queries->sdl-queries
```
hlship, line 216: I prefer a blank line between `def` forms.  There's a couple of cases of this.

hlship: :queries is such a hold over from early days Lacinia; we really should deprecate it, it causes problems.

That being said, should probably fold :queries in the Query object (likewise mutations and subscriptions) and then pretty print that.

It may be ok to skimp on some error checking, such as name collisions between Query fields an names in the :queries map ... incorrect SDL will be generated BUT that will be caught an instant later at schema compilation and/or calls to prevent-collision.

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +121,215 @@
...
+(defn ^:private edn-enums->sdl-enums
+  [enums]
+  (->> enums
+       (map (fn [[enum-name {values :values}]]
+              (str "enum " (name enum-name) "{\n" (->> values (map edn-enum-value->sdl-enum-value) (map name) (join "\n")) "\n}")))
+       (join "\n")))
+(defn ^:private edn-scalars->sdl-scalars
+  [scalars]
+  (->> (keys scalars)
+       (map name)
```
hlship, line 263: I think there's value to sorting by name in each of these blocks, for repeatability.

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +121,215 @@
...
+                   (->> locations
+                        (map directive-targets)
+                        (join " | ")))))
+       (join "\n")))
+
+(defn generate-sdl
+  "Translate the edn lacinia schema to the SDL schema."
+  [schema]
+  (->> schema
+       (map (fn [[key val]]
```
hlship, line 297: Again, repeatability; depending on the type of map (array-map vs. hash-map, etc.) this order of all this could shift dramatically; for small maps it's in order of keys added, in larger maps (hash-map) it's related to the hash of the key.

I'd say a good fixed order would be:

- directive defs
- scalars
- enums
- unions
- interfaces
- input-objects
- objects (including Query, Mutation, etc., rolled in)

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +121,215 @@
...
+                :objects (edn-objects->sdl-objects val)
+                :queries (edn-queries->sdl-queries val)
+                :interfaces (edn-interfaces->sdl-interfaces val)
+                :scalars (edn-scalars->sdl-scalars val)
+                :unions (edn-unions->sdl-unions val)
+                :input-objects (edn-input-objects->sdl-input-objects val)
+                :mutations (edn-mutations->sdl-mutations val)
+                :enums (edn-enums->sdl-enums val)
+                :directive-defs (edn-directive-defs->sdl-directives val)
+                :roots "")))
```
hlship, line 308: Roots could be important, though my experience with Apollo is that it's fragile if the service schemas don't agree on the names of the root objects.  That may have changed since I looked at it > 1 year ago.

hlship (commented):
Thanks for the initial effort!  I've outlined some improvements that could be made.

dev-resources/edn-federation.sdl
hmy3743: It's a bit complicated, but I'll try.

deps.edn
hmy3743: Removed match based on review.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: The input is edn, so there seems to be no problem. Can you give me an example?

src/com/walmartlabs/lacinia/federation.clj
hmy3743: Sorting has been added to reflect reviews.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: Sorting has been added to reflect reviews.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: Edited to reflect the review.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: Well, I don't understand what you mean, can you elaborate on that?

hlship: In the SDL you can override the default names of the Query, Mutation, and Subscription objects using the `schema` keyword (https://spec.graphql.org/October2021/#sec-Schema).

So this code must honor that, but must also (as necessary) emit the `schema` content into the output SDL.

hmy3743: Oh, I didn't know there was such a grammar. reflected.

hlship (commented):
As I understand the Apollo federation documentation, the @extends and other directive, scalar types, etc. are not explicit in the input SDL but still available, and should not be generated into the output SDL.  In the systems I've developed under federation, we start with SDL document and provided that document, as is ... you can see that in the code in master.

hmy3743:
> As I understand the Apollo federation documentation, the @extends and other directive, scalar types, etc. are not explicit in the input SDL but still available, and should not be generated into the output SDL. In the systems I've developed under federation, we start with SDL document and provided that document, as is ... you can see that in the code in master.

I am not sure if I understand the review comment well, but if you look at the [document](https://www.apollographql.com/docs/federation/federation-spec/#federation-schema-specification) about apollo's federation subgraph, it says that the server should provide some schemas. Then shouldn't the schema be exposed to the router through query{_service{sdl}}}?

hlship:
I'll work on resolving this against the Apollo documentation.

hlship:
Federation is a moving target, we're compatible with 1.0, but 2.0 changes things: https://www.apollographql.com/docs/federation/federated-types/federated-directives

In any case, I don't think the current version of this code will work because it includes the Federation (1.0) annotations and objects injected into the schema.

In the current version, based on reading SDL, you can see that those elements aren't in the input SDL (they are provided after that is parsed) and then we use the input SDL as-is when exposing the schema to the Apollo gateway.

I don't know exactly what would happen if we include those elements in the generated SDL but I suspect Apollo will reject the schema.

hlship:
I'm struggling to find time to come back and finish reviewing this.  I want to test what happens when the injected Fed 1.0 directives and types are _not_ filtered out when generating SDL from the schema (that is, see how Apollo reacts).

deps.edn
hmy3743: The core.match dependency is gone; the renderer uses ordinary conditionals. The refresh merges current upstream dependencies and uses the existing clojure.data.json dependency for string escaping. Full suite: 405 tests / 920 assertions, no failures or errors; focused clj-kondo: no warnings or errors.

dev-resources/edn-federation.sdl
hmy3743: The generated SDL now has stable section order, indentation, and sorted definitions/fields/arguments. Descriptions use escaped quoted strings so leading whitespace, line breaks, and triple quotes round-trip exactly. The updated fixture also passes Apollo composition.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: You were right: reading EDN does not make its strings safe to embed in SDL. Descriptions now use clojure.data.json/write-str; regression tests round-trip quotes, triple quotes, backslashes, newline, CR, tab, backspace, and form feed through the SDL parser.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: String values and argument descriptions now use JSON-compatible GraphQL escaping as well. The same control-character/backslash regression checks the parsed argument description and default value, not just generated text.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: The shorthand query/mutation/subscription fields are folded into their configured root objects before rendering. Added a regression with custom Read/Write/Watch roots and an existing Read field, and cleaned up the form spacing in the touched code.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: Every map-backed definition block, field list, argument list, and directive-argument map is sorted by name; directive location sets are also sorted. Tests compare differently ordered maps with more than eight fields, while preserving sequence order such as enum values.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: The renderer now visits an explicit fixed section list, with an optional schema block followed by directive definitions, scalars, enums, unions, interfaces, inputs, and objects. Automatic _service SDL generation removes foundation definitions; generic generate-sdl retains user-provided definitions. Field directives such as @external are retained. The Federation 1 fixture composes with a catalog subgraph using @apollo/composition 2.14.4.

src/com/walmartlabs/lacinia/federation.clj
hmy3743: Custom roots are emitted in a schema block and honored when folding all three operation shorthands. The Read/Write/Watch regression parses the output and verifies both the root mapping and merged fields. The federation fixture now declares only its real Query root.

src/com/walmartlabs/lacinia/federation.clj
```diff
@@ -119,25 +120,269 @@
...
+
+(defn ^:private edn-type->sdl-type
+  [type]
+  (if (seq? type)
+    (let [[hd & tl] type]
+      (cond
+        (nil? hd) ""
+        (= 'non-null hd) (str (apply-list edn-type->sdl-type tl) "!")
+        (= 'list hd) (str "[" (apply-list edn-type->sdl-type tl) "]")
+        (= 'String hd) "String"
```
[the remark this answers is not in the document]
hmy3743: The redundant built-in scalar cases were removed; symbol types are handled by the generic symbol branch. The current parser round-trip and full test suite pass.

PR #421, opened 2022-08-16
Make ArgumentDef protocol support Directives

happy-river:
Add support for `Directives` to `ArgumentDef` so that field resolver functions can navigate from the `FieldSelection` instance returned by `com.walmartlabs.lacinia.executor/selection` to any directives placed on their arguments in the schema.


hlship:
Thanks for filling this gap and for providing a test.

PR #425, opened 2022-09-07
Add a compilation option to apply field directives to streamer functions

happy-river:
Resolver functions can be preprocessed during compilation of the schema using the option `:apply-field-directives`.  Add a new option `:apply-subscription-field-directives` to allow streamer functions to be preprocessed in the same way.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 [Closed Issues](https://github.com/walmartlabs/lacinia/issues?q=is%3Aclosed+milestone%3A1.2)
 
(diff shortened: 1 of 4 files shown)


hlship:
Have you signed the CLA?  Seems like that check is broken.

happy-river:
> Have you signed the CLA? Seems like that check is broken.

Yes, see #421.

hlship:
Sorry to bug you about that, it used to indicate that as a check but it's now inconsistent and I can't get help from the open source team at Walmart to fix it.

hlship (approved):

PR #433, opened 2022-12-02
Do not coerce nil into empty vector when an argument type is list and null is given

ljb7977:
Hi team!

Currently, when literal `null` is passed as an input to a `List` type argument, it gets coerced into an empty vector, which is incorrect. 

```clojure
(let [schema (schema/compile {:queries
                                {:echo
                                 {:type :String
                                  :args {:input {:type '(list (non-null Int))}}
                                  :resolve (fn [_ args _]
                                             (str "Echo: " (:input args)))}}})]
    (execute schema "query { echo(input: null) }" nil nil))
=> {:data #ordered/map([:echo "Echo: []"])}
```
(The example format is borrowed from @duddlf23 's one :) )

However, according to https://spec.graphql.org/October2021/#sec-List.Input-Coercion, the input value should be coerced to list only if it is "not a list and not a null". So the correct result of the coercion should be `nil` in this case.

This is because the `mapv` function returns an empty vector when its input is `nil`. In the body of the multimethod `process-literal-argument :array`, `mapv` is called even if the arg-value is `nil`.  https://github.com/walmartlabs/lacinia/blob/3f3ca1eb5b4e8baef48ad56ab661739cda0dd8fb/src/com/walmartlabs/lacinia/parser.clj#L363

This PR fixes `process-literal-argument` to return `nil` by not passing the `arg-value` to `mapv` if it is `nil`.

Co-authored-by: Hyunwoo Nam <namenu@gmail.com>

### DIFF
--- a/src/com/walmartlabs/lacinia/parser.clj
+++ b/src/com/walmartlabs/lacinia/parser.clj
@@ -359,8 +359,10 @@
       (throw-exception "Provided argument value is an array, but the argument is not a list.")
 
       :list
-      (let [fake-argument-def (use-nested-type argument-definition)]
-        (mapv #(process-literal-argument schema fake-argument-def %) arg-value)))))
+      ;; if arg-value is nil, should return nil (no coercion is required)
+      (when (some? arg-value)
+        (let [fake-argument-def (use-nested-type argument-definition)]
+          (mapv #(process-literal-argument schema fake-argument-def %) arg-value))))))
 
 (defn ^:private decapitalize
   [s]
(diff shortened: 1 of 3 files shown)


hlship:
Could you provide a test case here?

ljb7977:
Sure. Will add them today or tomorrow :)

hlship:
Did you run all tests?  I'm seeing some failures, but haven't had a chance to investigate whether the failed tests are valid or invalid.

ljb7977:
Sorry I missed the failing tests. 

In the last commit added, I changed the expected results of those existing tests to meet the specification of input coercion for the null value.
It seems that the spec changed in October 2016. In the [spec document as of Oct 2016](https://spec.graphql.org/October2016/#sec-Lists), the statement 
"`If the value passed as an input to a list type is not a list and not the null value, it should be coerced as though the input was a list of size one`" appears for the first time. So I think the expected results of those tests need to change.

hlship:
Hello, thanks for this patch. I apologize if I was overly terse, I'm both busy and sick.  I'm about to do a final review and merge.

hlship (approved):
I really appreciate you going to the effort of tying the changes back to the GraphQL spec.

PR #443, opened 2023-04-19
Add support for @deprecated directives on :argument-definition, :input-field-definition

sundbry:
Hi Walmart Labs,

While the @deprecated annotation is suported for enum values and type fields, it was not supported for input field definitions. GraphQL @deprecated directives are also valid on Input types in the most recent [Working Draft of the GraphQL Spec](http://spec.graphql.org/draft/#sec--deprecated).


hlship (approved):

hlship:
Many hands make light work - thanks!

PR #449, opened 2023-07-21
Fix runtime error when aggregating results with error-containing left-value

namenu:
The query below causes an infinite wait when executed.

```graphql
query HangingQuery($id: ID!) {
  node(id: $id) {
    ... on Post {
      ... on Post {
        author {
          alwaysError
        }
      }
      author {
        name
      }
    }
  }
}
```

On the other hand, if you reverse the order of the field declarations, it will not.

```graphql
query HangingQuery($id: ID!) {
  node(id: $id) {
    ... on Post {
      author {
        name
      }
      ... on Post {      # change the order
        author {
          alwaysError
        }
      }
    }
  }
}
```

It's worth noting that the fragment must be used to reproduce the error.

I guess in the process of doing `merge-selected-values`,
even though the left-value value is nil (which is resolved by alwaysError),
`update` to nil is called (which cause runtime error).

This PR fixes this issue.

### DIFF
--- a/src/com/walmartlabs/lacinia/executor.clj
+++ b/src/com/walmartlabs/lacinia/executor.clj
@@ -181,9 +181,16 @@
   is an ResultTuple."
   [left-value right-value]
   (if (su/is-result-tuple? right-value)
-    (let [{:keys [alias value]} right-value]
-      (if (contains? left-value alias)
+    (let [{:keys [alias value]} right-value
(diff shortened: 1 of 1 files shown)


hlship:
I'm sorry that I haven't kept track of this.  I'd like to create a 1.2.2 release, but looks like your patch will need to be updated.  If you can put that together, I'd be happy to merge.

PR #462, opened 2024-06-26
Fixes a runtime error that occurs when deep-merging fragments

namenu:
Fixes a bug where inconsistent results occur based on the fragment's declared order.

I fixed a similar issue in #453, but in that case the fragment was nested, whereas this case is not.

This fails:
```graphql
query MyQuery {
  node(id: "1000") {
    ... on Post {
      id
      ...PostFragment
    }
  }
}

fragment PostFragment on Post {
  author {
    alwaysFail
  }
  ...PostFragment2   # <--
}

fragment PostFragment2 on Post {
  author {
    name
  }
}
```

While this is not:
```graphql
query MyQuery {
  node(id: "1000") {
    ... on Post {
      id
      ...PostFragment
    }
  }
}

fragment PostFragment on Post {
  ...PostFragment2   # <--
  author {
    alwaysFail
  }
}

fragment PostFragment2 on Post {
  author {
    name
  }
}
```

My guess is that `deep-merge` doesn't handle the situation where nil comes in instead of map.

### DIFF
--- a/src/com/walmartlabs/lacinia/internal_utils.clj
+++ b/src/com/walmartlabs/lacinia/internal_utils.clj
@@ -409,7 +409,8 @@
 
 (defn- null?
   [v]
-  (= v :com.walmartlabs.lacinia.schema/null))
+  (or (nil? v)
+      (= v :com.walmartlabs.lacinia.schema/null)))
 
 (defn deep-merge
   "Merges two maps together.  Later map override earlier.
--- a/src/com/walmartlabs/lacinia/internal_utils.clj
+++ b/src/com/walmartlabs/lacinia/internal_utils.clj
@@ -407,11 +407,18 @@
            nil
           coll))
 
+(defn- null?
+  [v]
+  (= v :com.walmartlabs.lacinia.schema/null))
+
 (defn deep-merge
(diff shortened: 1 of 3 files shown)


src/com/walmartlabs/lacinia/internal_utils.clj
```diff
@@ -407,11 +407,19 @@
...
+  [v]
+  (or (nil? v)
+      (= v :com.walmartlabs.lacinia.schema/null)))
+
 (defn deep-merge
   "Merges two maps together.  Later map override earlier.
   If a key is sequential, then each element in the list is merged."
   [left right]
   (cond
+    (or (null? left) (null? right))
```
namenu, line 420: I observed both cases that left/rigt can be `::null` or `nil.
It depends on whether the object that catches bubbled-up error is non-null or not.

Test case in L169 covers this.

namenu:
@hlship Thanks. We have encountered a regression bug after applying this changes. I am not confident much so that I can handle this problem since it requires profound understanding of the internal design.
I'm still finding the way how to reproduce the bug. If you have any idea about this please let me know.

namenu:
Reopened #463

